Issue #294, opened 2026-03-27
bug: 2-hop reverse arrow pattern (a)-[:REL]->(b)<-[:REL]-(c) returns 0 rows


ryaker:
## Summary

The mutual neighbor / co-actor pattern `(a)-[:REL]->(b)<-[:REL]-(c)` returns 0 rows when it should return results. This is a correctness bug, not just a performance issue.

## Reproduction

Discovered via Neo4j Movies compatibility tests in SparrowTest:

```cypher
-- Find co-actors of Tom Hanks (people who acted in the same movie)
MATCH (tom:Person {name: "Tom Hanks"})-[:ACTED_IN]->(m:Movie)<-[:ACTED_IN]-(coactor:Person)
RETURN coactor.name
-- Expected: multiple co-actors
-- Actual: 0 rows
```

The same pattern fails across datasets:
- Movies: co-actor query returns 0
- Facebook benchmark Q8 (mutual neighbors): functionally broken
- Any `(a)-[:R]->(m)<-[:R]-(b)` pattern

## Workaround

The 4-hop variant with alternating forward arrows works:
```cypher
MATCH (tom:Person {name: "Tom Hanks"})-[:ACTED_IN]->(m:Movie)
MATCH (coactor:Person)-[:ACTED_IN]->(m)
WHERE coactor.name <> "Tom Hanks"
RETURN coactor.name
```

## Related Issues

- #287 (perf: HashSet intersection for mutual neighbors) — the performance issue assumes the pattern works; this bug means it doesn't work at all
- #282 (2-hop COUNT aggregate) — similar 2-hop traversal area

## Root Cause (suspected)

The backward CSR lookup in `hop.rs` for the `<-` direction in the second hop may not be resolving correctly, or the `second_hop_incoming` flag logic doesn't handle this specific pattern.

## Impact

This blocks:
- Co-actor / mutual friend queries (common graph pattern)
- Any bidirectional 2-hop traversal
- Q8 benchmark accuracy (currently returns wrong results, not just slow)

PR #359, opened 2026-03-29
perf(engine): property-index fast path for variable-length source node (Q5)

ryaker:
## Summary

- Adds a property-index fast path in `execute_variable_length` (`path.rs`) before the `for src_slot in 0..hwm_src` source scan loop
- When the source node pattern has a single indexed inline prop (e.g. `{name: 'Alice'}`), `try_index_lookup_for_props` resolves it to 1 slot in O(1) — eliminating the full label scan
- Falls back to `0..hwm_src` full scan when the index cannot resolve the filter (multi-prop, non-literal, or un-indexed label)

## Root cause

```rust
// Before: always scans ALL source slots even for single-node queries
for src_slot in 0..hwm_src {
    let src_props = read_node_props(...);         // disk read per slot
    if !self.matches_prop_filter(...) { continue; } // filter applied AFTER read
    // BFS from this source...
}
```

For `MATCH (a:Person {name:'Alice'})-[:KNOWS*1..3]->(f)` on a 4K-node graph, this reads 4K property entries just to find the one starting node.

```rust
// After: resolve source via index first, fall back to full scan
let resolved_slots: Option<Vec<u64>> = {
    let prop_idx = self.prop_index.borrow();
    try_index_lookup_for_props(&src_node_pat.props, src_label_id, &prop_idx)
        .map(|slots| slots.into_iter().map(|s| s as u64).filter(|&s| s < hwm_src).collect())
};
let src_iter: Box<dyn Iterator<Item = u64>> = match resolved_slots {
    Some(slots) => Box::new(slots.into_iter()),
    None => Box::new(0..hwm_src),
};
for src_slot in src_iter { ... }
```

## Investigation findings

- Q5 variable-length BFS is NOT a Phase 4 regression — the source scan was always O(hwm_src) per query
- `delta_all` and `delta_idx` (the CSR edge index) were already hoisted outside the loop in SPA-275/SPA-283; the remaining bottleneck is source slot enumeration
- The "7K baseline" ops rate was pre-#294 wrong results at faster apparent speed
- `build_delta_index` is already called once per query, not per source slot (already fixed)

## Test plan

- [ ] `cargo test -p sparrowdb-execution` — all pipeline tests pass
- [ ] `cargo test -p sparrowdb` — all integration tests pass
- [ ] Benchmark Q5 after merge: queries with single-source inline props on indexed labels expected to improve significantly; multi-source / un-indexed queries unchanged

### DIFF
--- a/crates/sparrowdb-execution/src/engine/path.rs
+++ b/crates/sparrowdb-execution/src/engine/path.rs
@@ -423,14 +423,13 @@ impl Engine {
         // calls inside the loop (check_deadline, execute_variable_hops, etc.).
         let resolved_slots: Option<Vec<u64>> = {
             let prop_idx = self.prop_index.borrow();
-            try_index_lookup_for_props(&src_node_pat.props, src_label_id, &prop_idx)
-                .map(|slots| {
-                    slots
-                        .into_iter()
-                        .map(|s| s as u64)
-                        .filter(|&s| s < hwm_src)
-                        .collect()
-                })
+            try_index_lookup_for_props(&src_node_pat.props, src_label_id, &prop_idx).map(|slots| {
+                slots
+                    .into_iter()
+                    .map(|s| s as u64)
+                    .filter(|&s| s < hwm_src)
+                    .collect()
+            })
         };
         let src_iter: Box<dyn Iterator<Item = u64>> = match resolved_slots {
             Some(slots) => Box::new(slots.into_iter()),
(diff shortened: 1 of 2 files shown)


ryaker:
## Benchmark results

Criterion measurements (1K-node / 20K-edge synthetic graph, isolated sequential runs):

| Branch | Q5 variable-path p50 |
|--------|----------------------|
| main   | 30.5ms               |
| fix/q5 | 28.4ms (within noise, p=0.37) |

**Result: no measurable improvement on this graph size.** The BFS cost (single source visiting most of 1K-node graph at depth 1..3) dominates, and the source-resolution overhead (999 property reads from page-cached 8KB column file) is negligible at ~50µs. The fix is expected to show improvement when:
- Many source nodes AND cold column files (first read from disk, e.g. post-restart), where 4K slot reads × SSD latency = measurable overhead
- Large graphs where source scan is a non-trivial fraction of total query time

No regression detected. Algorithmic correctness confirmed. Ships as a forward-looking optimization.

PR #411, opened 2026-06-10
feat(cypher): UNWIND MATCH SET/DELETE + DELETE RETURN (SPA-415)

evuk79:
Adds two new Cypher capabilities to SparrowDB:

1. UNWIND ... MATCH ... SET/DELETE [RETURN ...]
   - New AST variant: UnwindMatchMutateStatement
   - Parser: parse_unwind() extended to handle MATCH + mutation tail
   - Binder: binds MATCH patterns, UNWIND expr uses
   - Engine: is_mutation() returns true for UnwindMatchMutate
   - db.rs: execute_unwind_match_mutate_with_params() evaluates
     the UNWIND list and applies mutations per-row within a single
     write transaction (eliminates N+1 query parse overhead)

2. DELETE ... RETURN count(n)
   - MatchMutateStatement gains optional return_clause field
   - Parser: parse_match_or_match_mutate() parses RETURN after SET/DELETE
   - db.rs: build_mutate_return() helper returns mutation count
   - All three mutation paths (direct, deadline, with-params) updated

These capabilities enable awknow-rs to replace N indidvidual
MATCH...SET queries with a single UNWIND query, and replace
COUNT-then-DELETE with a single DELETE...RETURN query.

Closes the N+1 and COUNT+DELETE anti-patterns in SparrowGraphAccess.

## Summary by CodeRabbit

* **New Features**
  * MATCH ... SET/DELETE now accepts an optional RETURN clause so mutations can return results (e.g., mutation counts).
  * UNWIND ... AS ... MATCH ... SET/DELETE added to batch mutations over list elements, with optional RETURN to aggregate results across the batch; supports using element fields in MATCH/SET expressions.

### DIFF
--- a/crates/sparrowdb-cypher/src/ast.rs
+++ b/crates/sparrowdb-cypher/src/ast.rs
@@ -327,6 +327,34 @@ pub struct MatchMutateStatement {
     /// For SET, there may be multiple items (comma-separated).
     /// For DELETE, this always contains exactly one `Mutation::Delete`.
     pub mutations: Vec<Mutation>,
+    /// Optional RETURN clause after mutations (e.g. `DELETE n RETURN count(n)`).
+    pub return_clause: Option<ReturnClause>,
+}
+
+/// UNWIND … MATCH … SET/DELETE statement (SPA-415).
+///
+/// Iterates a list expression, binding each element to `alias`, then for
+/// each element executes a MATCH + mutation (SET or DELETE).
+///
+/// ```cypher
+/// UNWIND $updates AS row
+/// MATCH (n:Memory {id: row.id})
+/// SET n.score = row.score
+/// ```
+#[derive(Debug, Clone, PartialEq)]
+pub struct UnwindMatchMutateStatement {
+    /// The list expression to unwind (e.g. `$updates`).
+    pub expr: Expr,
+    /// Variable bound to each list element (e.g. `row`).
+    pub alias: String,
+    /// The MATCH patterns whose nodes are mutated.
+    pub match_patterns: Vec<PathPattern>,
+    /// Optional WHERE predicate on the MATCH.
+    pub where_clause: Option<Expr>,
+    /// The mutations to apply (SET or DELETE).
+    pub mutations: Vec<Mutation>,
+    /// Optional RETURN clause after mutations.
+    pub return_clause: Option<ReturnClause>,
 }
 
 /// A single projection in a WITH clause: `expr AS alias`.
@@ -501,6 +529,8 @@ pub enum Statement {
     /// MATCH … MERGE (a)-[r:TYPE]->(b) — find-or-create a relationship (SPA-233).
     MatchMergeRel(MatchMergeRelStatement),
     MatchMutate(MatchMutateStatement),
+    /// UNWIND … MATCH … SET/DELETE — batched mutations via list parameter (SPA-415).
+    UnwindMatchMutate(UnwindMatchMutateStatement),
     OptionalMatch(OptionalMatchStatement),
     MatchOptionalMatch(MatchOptionalMatchStatement),
     Union(UnionStatement),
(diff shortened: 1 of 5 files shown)


ryaker:
Status update — thanks for this, and apologies it sat unreviewed for two months.

**The six unresolved review threads are already satisfied by your code.** All five high-priority Gemini findings about `params` being evaluated then discarded were fixed in `8a08721`; `params` is threaded through `execute_unwind_mutate_inner`, `engine.with_params()`, and `resolve_set_value`. The threads were just never resolved, and Gemini has since been retired so nothing was going to clear them. That was never the blocker.

I rebased the branch onto current `main` (one import-list conflict, plus two unused imports the rebase exposed) and wrote the e2e coverage the PR was missing — `spa_415_unwind_match_mutate.rs`, 15 tests covering the literal and `$param` paths, `SET n.prop = row.field`, DELETE and DETACH DELETE, `DELETE … RETURN count()`, empty/no-match elements, durability across a reopen, and the same-snapshot visibility semantic your doc comment documents. 14 pass.

**Two things block merging, both filed:**

**#468 — this PR regresses `UNWIND … MATCH … RETURN`.** `spa_237_unwind_match` passes 6/6 on `main` and 4/6 on this branch, bisected to `9600c0e`. `parse_unwind`'s new dispatch pre-consumes the MATCH and passes it as `leading_match` while also setting `leading_unwind`; `execute_pipeline_inner` tests them in an `if/else if` with `leading_unwind` first, so the MATCH is silently discarded. Results come back `["null", "null"]` with wrong row counts. That existing test file would have caught it immediately — it's the strongest argument for the tests this PR needed.

**#467 — `MATCH (n:L {p: row})` with a bare alias matches every node.** `resolve_pattern_props` only resolves `Expr::PropAccess`, so a bare `Expr::Var` reaches `eval_expr`, evaluates to `Null`, and `matches_prop_filter_static` treats a Null filter as always-passing. On a `SET` that writes to every node of the label. Verified: 3 nodes, 1 named, all 3 mutated. The underlying fail-open isn't yours — your branch just supplies a reachable path to it — so I filed it separately for a wider audit.

Neither is a design objection. The feature and the params plumbing look right; #468 is a narrow parser/pipeline interaction and #467 is pre-existing. I'll pick both up unless you'd rather take them.

One logistics note: this PR is from `evuk79:main`. Rather than force-push a rebase onto your branch, I'll land the work from our side with your authorship preserved on the commits and credit in the merge, unless you'd prefer I push here instead — your call.

ryaker:
Opened #470 with your work rebased onto current `main`, your four commits preserved with original authorship, plus the test coverage and the two regression fixes described above.

I went this route rather than force-pushing a rebase onto `evuk79:main` — `maintainer_can_modify` is set, so I could have, but rewriting a contributor's own branch isn't something I'll do without you asking. If you'd rather own the branch, say so and I'll close #470 and push there instead.

Keeping this open until #470 merges, at which point I'll close it as superseded with credit to you. The feature and the params plumbing are yours; what was missing was tests, and those turned out to matter — running the pre-existing `spa_237_unwind_match` is what caught #468.

ryaker:
Superseded by #470, which merged and shipped in **0.1.24**.

Your four commits are preserved in main with original authorship — the feature and the params plumbing landed as you wrote them. What #470 added was the test coverage (16 e2e tests) and fixes for two regressions the feature introduced that no existing test caught:

- **#468** — `parse_unwind`'s lookahead passed the pattern as `leading_match` while `leading_unwind` was also set. `execute_pipeline_inner` picks the leading clause with an `if/else-if` testing `leading_unwind` first, so the MATCH was silently dropped and `UNWIND … MATCH … RETURN` returned `["null", "null"]`. Also, dispatching on `Token::Where` routed reads carrying a predicate into the mutation tail, which rejected them.
- A silent HNSW data-loss regression: `UNWIND … SET n.embedding = $emb` wrote the property and skipped vector-index maintenance entirely, so the vector was invisible to search forever while the write reported success.

Neither was your doing — the first was only findable by running the pre-existing `spa_237_unwind_match` suite against the branch, which is what caught it. Thanks for the contribution; `UNWIND … MATCH … SET/DELETE` and `DELETE … RETURN count()` are live on npm now.

PR #414, opened 2026-08-01
fix(clippy): resolve three lints failing on clean main under Rust 1.97

ryaker:
## Problem

`cargo clippy --all-targets -- -D warnings` **fails on unmodified `main`** with clippy 0.1.97 (2026-07-14). CI uses `dtolnay/rust-toolchain@stable`, so the clippy step fails for every PR regardless of its contents — currently blocking #412 and #413.

Recurring pattern: newer clippy flags code that passed under the version in use when it was written.

## Changes

Three lints, all mechanical and behavior-preserving:

| File | Lint | Change |
|---|---|---|
| `sparrowdb-execution/src/engine/expr.rs:60` | `iter_kv_map` | `for (_k, v) in vals.iter()` → `for v in vals.values()` |
| `sparrowdb-cli/src/main.rs:171` | `question_mark` | `match db.execute(..) { Ok(..) =>, Err(e) => return Err(e.into()) }` → `?` |
| `sparrowdb/tests/spa_fulltext.rs:217,233` | `needless_borrow` | redundant `&` in `assert!`/`assert_eq!` format arguments |

## Verification

- `cargo clippy --workspace --exclude sparrowdb-ruby --all-targets -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean
- `cargo test -p sparrowdb-execution -p sparrowdb-cli` — all green
- `cargo test -p sparrowdb --test spa_fulltext` — 5 passed

`sparrowdb-ruby` is excluded locally because `magnus`/`rb_sys` doesn't compile against the installed Ruby (~79 errors, pre-existing). CI's clippy step runs in the `test` job without a Ruby setup step, so if that crate has its own lints they'll surface there — none of the three fixed here are in it.

## Note

Found while verifying 0.1.22 for release. Kept separate from #412 (CI pipefail) and #413 (duplicate-edge bug) since it's unrelated to both — but it needs to land for either to show green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for query execution, ensuring failures are reported consistently.
  * Preserved full-text search behavior while improving internal result processing.

* **Tests**
  * Updated full-text search checks to more accurately validate returned values.

### DIFF
--- a/crates/sparrowdb-cli/src/main.rs
+++ b/crates/sparrowdb-cli/src/main.rs
@@ -168,13 +168,9 @@ fn main() {
 
 fn cmd_query(db_path: &std::path::Path, cypher: &str) -> Result<(), Box<dyn std::error::Error>> {
     let db = sparrowdb::GraphDb::open(db_path)?;
-    match db.execute(cypher) {
-        Ok(result) => {
-            let json = query_result_to_json(&result);
-            println!("{}", serde_json::to_string_pretty(&json)?);
-        }
-        Err(e) => return Err(e.into()),
-    }
+    let result = db.execute(cypher)?;
+    let json = query_result_to_json(&result);
+    println!("{}", serde_json::to_string_pretty(&json)?);
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


ryaker:
@coderabbitai review

Issue #299, opened 2026-03-27
arch: Chunked vectorized query execution pipeline (TuringDB-style)


ryaker:
## Technical Spec: Chunked Vectorized Query Execution Pipeline

**Issue**: #299  
**Status**: Spec complete, ready for review  
**Estimated Complexity**: XL (multi-quarter, 4 phases)  
**Dependencies**: #283 (delta index) — DONE, #285/#288 (batch property reads) — DONE

---

## 1. Problem Statement

SparrowDB's query engine processes graph traversals **node-at-a-time**: each source node triggers a separate CSR offset lookup, delta-log merge, property decode, and WHERE evaluation. This is fundamentally at odds with modern CPU architecture:

**Current hot path in `hop.rs` (1-hop, simplified):**
```
for each src_slot in scan(label):
    csr_nb = csr.neighbors(src_slot)           // 1 offset lookup + slice
    delta_nb = delta_index.get(src_slot)        // 1 hash lookup
    all_nb = csr_nb ++ delta_nb                 // allocate + copy
    for each dst_slot in all_nb:
        props = store.read_node_props(dst_slot) // 1 column-file seek per property
        vals = decode(props)                    // 1 decode per value
        if eval_where(vals): emit row           // 1 HashMap allocation per row
```

**Why this is slow:**
1. **Cache hostility**: Each `csr.neighbors(slot)` indexes into an offset array at an unpredictable location. Adjacent source slots may have neighbors scattered across the neighbor array, but the CPU prefetcher cannot anticipate the access pattern because offsets are data-dependent.
2. **Allocation pressure**: Every row builds a `HashMap<String, Value>` for WHERE evaluation and projection. For a 2-hop query with 22 avg degree, that is 484+ HashMap allocations per source node.
3. **No amortization**: Property column files are mmap'd, but reads are per-slot. Reading props for slots [5, 12, 47, 103] requires 4 separate offset calculations. `batch_read_node_props` (SPA-285) helps but is only used in specific code paths, not as the universal data-access pattern.
4. **No pipeline breaks**: The entire result set materializes in memory before ORDER BY / LIMIT can discard rows. A `LIMIT 10` on a 2-hop query still expands all 484 neighbors before sorting.

**Performance gap** (LDBC SNB, SF-01, single-threaded):

| Query | SparrowDB | TuringDB-class | Gap |
|-------|-----------|----------------|-----|
| IC3 1-hop + filter | ~42ms | ~0.6ms | 70x |
[...]

PR #416, opened 2026-08-01
fix(engine): chunked pipeline bypassed multi-label MATCH (closes #415)

### LINKED ISSUE #289: feat(parser): multi-label nodes (n:A:B) — standard Cypher semantics (SPA-200)
### Problem

SparrowDB currently enforces a single label per node. Standard openCypher supports multi-label nodes: `CREATE (n:Person:Employee {name: "Alice"})` assigns both `:Person` and `:Employee` to the same node, and `MATCH (n:Person)` must return that node even though it also carries `:Employee`.

The current implementation hard-codes a 1:1 relationship between a node and its label at every layer:

| Layer | Current constraint |
|---|---|
| **NodeId encoding** | `(label_id: u32) << 32 \| slot: u32` — the label is physically embedded in the node ID |
| **Storage** | Column files live under `nodes/{label_id}/` — one directory per label |
| **Catalog** | `LabelId = u16`, simple name→id map, no multi-label association |
| **Execution** | `node.labels.first()` everywhere — only the first label is resolved |
| **Property index** | Keyed by `(label_id, col_id)` — assumes one label per node |

The parser already accepts `(n:A:B)` and populates `NodePattern.labels: Vec<String>`, but the engine silently drops all labels after the first.

---

### Design Decision: Primary Label with Secondary Label Set

After evaluating three options (see appendix), the recommended approach is **primary label + secondary label set**. This preserves the existing NodeId encoding and columnar storage layout while adding multi-label semantics on top.

#### Core Idea

- Every node retains exactly **one primary label** that determines its `NodeId` encoding and physical storage location (the `nodes/{label_id}/` directory). This is backward compatible.
- Additional labels are tracked in a **label-set side table** — a new catalog-level structure that maps `NodeId → Set<LabelId>`.
- MATCH resolution uses the side table to answer "does this node have label X?" without changing the NodeId packing.

---

#### Option A: Bitmap column per label-group (REJECTED)

Store a bitmap column (`col_LABELS.bin`) in each label directory. Each bit position maps to a `LabelId`. Pros: fast bitwise checks. Cons: wastes space (u16 label space = 8KB per node), complex resizing, doesn't scale past 64 labels without multi-word bitmaps.

#### Option B: Label-set table (RECOMMENDED)

A new file per primary label directory:

```
nodes/{primary_label_id}/secondary_labels.bin
```

Format: for each slot, a variable-length list of secondary `LabelId` values. Two sub-options:

- **B1 — Fixed-width bitset column**: A single `u64` column where each bit maps to a secondary label. Supports up to 64 total labels (sufficient for most graphs). Simple, cache-friendly, O(1) lookup.
- **B2 — Overflow file with offsets**: An offset column (`col_label_offsets.bin`) pointing into a packed array of `LabelId` values. Supports unlimited labels but adds indirection.

**Recommendation: B1 (fixed-width bitset)** for Phase 1. 64 labels covers the vast majority of use cases. If needed, B2 can be added later behind a feature flag.

#### Option C: Multi-label-id mapping table (REJECTED)

A global `node_labels` table: `(NodeId, LabelId)` with one row per node-label pair. Resembles a relational join table. Pros: simple model. Cons: requires a join on every scan, cannot leverage columnar locality, poor cache behavior.

---

### Parser Changes

**No changes needed.** The parser already handles `(n:A:B:C)` correctly:

```rust
// parser.rs:1451-1456
let mut labels = Vec::new();
while matches!(self.peek(), Token::Colon) {
    self.advance();
    let label = self.expect_label_or_type()?;
    labels.push(label);
}
```

The AST `NodePattern.labels` is already a `Vec<String>`. The only gap is that downstream consumers ignore everything after `labels[0]`.

---

### Catalog Changes

1. **No change to `LabelEntry` or `LabelId`** — each label name still maps to a unique `LabelId`.

2. **New TLV entry type: `NodeLabelSet`** — persists secondary label assignments:
   ```
   Tag: 0x03 (new)
   Fields: primary_label_id: u16, slot: u32, secondary_label_ids: Vec<u16>
   ```
   Appended to `catalog.tlv` when a multi-label node is created or when `SET n:NewLabel` is executed.

3. **In-memory index**: `HashMap<NodeId, HashSet<LabelId>>` loaded on catalog open. This is the authoritative source for "which labels does this node have?"

4. **Reverse index**: `HashMap<LabelId, HashSet<NodeId>>` — "which nodes have this label?" Used by MATCH to find nodes by secondary label without full scans.

---

### MATCH Semantics

`MATCH (n:A)` must return all nodes whose label set contains `:A`, regardless of whether `:A` is the primary or a secondary label.

#### Single-label query: `MATCH (n:A)`

1. If `:A` is the node's primary label → found via existing scan (no change).
2. If `:A` is a secondary label → found via the reverse index `secondary_label_index[A]`.
3. Union both result sets, deduplicate by `NodeId`.

#### Multi-label query: `MATCH (n:A:B)`

1. For each label in the pattern, collect the set of `NodeId`s that have that label.
2. Intersect all sets.
3. Return the intersection.

#### Implementation in `scan.rs`

Every call site that currently does:
```rust
let label = node.labels.first().cloned().unwrap_or_default();
```
must be updated to:
- Resolve **all** labels in the pattern.
- If `labels.len() == 1`, use the existing primary-label scan path, plus union with the reverse index for that label as a secondary.
- If `labels.len() > 1`, intersect across all labels.

The `__labels__` metadata injected for `labels(n)` must return the full label set, not just the primary label.

---

### CREATE Semantics

`CREATE (n:Person:Employee {name: "Alice"})`:

1. The **first label** (`:Person`) is the primary label — determines `NodeId` encoding and storage directory.
2. All subsequent labels (`:Employee`) are secondary labels.
3. After `create_node(label_id, props)` succeeds:
   - Write a `NodeLabelSet` TLV entry to the catalog.
   - Update the in-memory label-set index and reverse index.
4. The `labels()` function returns `["Person", "Employee"]`.

**Open question**: Should primary label selection be deterministic (alphabetical) or positional (first in syntax)? Recommendation: **positional** — matches user intent and is consistent with Neo4j behavior.

---

### SET / REMOVE Label Support (Future)

Standard Cypher supports `SET n:NewLabel` and `REMOVE n:OldLabel`. This spec does not cover these mutations but the design accommodates them:

- `SET n:X` → append `X` to the node's label set in the catalog side table.
- `REMOVE n:X` → remove `X` from the set. Removing the primary label is an error (or requires a migration to a new NodeId, which is a much larger change).

---

### Property Index Impact

The `PropertyIndex` is keyed by `(label_id, col_id)`. A node created as `(:Person:Employee {name: "Alice"})` stores its columns under the primary `label_id` (Person).

**Implication**: `CREATE INDEX ON :Employee(name)` will NOT automatically index the node above, because its data lives in `:Person`'s column directory.

**Options**:

1. **Cross-label index awareness** (Phase 2): When building the index for `(:Employee, name)`, also scan nodes whose secondary labels include `:Employee`. This requires iterating the reverse index.
2. **Phase 1 limitation**: Document that property indexes apply to primary labels only. `CREATE INDEX ON :Person(name)` works; a secondary-label index query falls back to a full scan + filter.

**Recommendation**: Phase 1 ships with the limitation documented. Phase 2 adds cross-label index support.

---

### `labels()` Function Changes

Currently `labels(n)` returns `[primary_label]`. After this change:

- `labels(n)` returns the full ordered label set: `[primary_label, secondary_1, secondary_2, ...]`.
- Primary label is always first (stable ordering for deterministic results).
- The `__labels__` metadata key injected in `scan.rs` must be populated from the catalog side table.

---

### Migration

**Existing single-label databases require zero migration.** The design is purely additive:

- Nodes with no secondary labels have no entry in the `NodeLabelSet` TLV section or the side table.
- The reverse index for single-label nodes is empty.
- All existing queries continue to work identically.
- The `catalog.tlv` format is append-only; new entry types are ignored by older readers (forward-compatible by design).

---

### MERGE Semantics

`MERGE (n:Person:Employee {name: "Alice"})` currently requires a single label. Changes needed:

1. Parser: `parse_merge` must accept multiple labels (currently errors after the first colon-label).
2. MERGE match phase: look up nodes that have **all** specified labels plus matching properties.
3. MERGE create phase: if no match, create with all labels (same as CREATE semantics above).

---

### Risks & Open Questions

| # | Item | Status |
|---|---|---|
| 1 | **NodeId stability**: A node's primary label is baked into its NodeId. Changing primary labels post-creation requires re-encoding the NodeId, which breaks all edge references. | Accept: primary label is immutable. |
| 2 | **Label count limit**: The bitset approach (B1) caps total distinct labels at 64. Is this sufficient? | Likely yes for Phase 1. Monitor and add overflow (B2) if needed. |
| 3 | **Property index gap**: Secondary-label indexes won't cover cross-label nodes in Phase 1. | Document limitation. Phase 2 follow-up. |
| 4 | **Edge storage**: Edges reference NodeIds. Since NodeId encoding is unchanged, no edge migration is needed. | Confirmed safe. |
| 5 | **WAL replay**: The WAL codec must learn to replay `NodeLabelSet` entries. | Needs new WAL op code. |
| 6 | **`REMOVE n:PrimaryLabel`**: Should this be an error or trigger a node migration? | Recommend: error in Phase 1. |
| 7 | **Ordering of `labels()`**: Neo4j returns labels in alphabetical order. Should we match? | Recommend: insertion order (primary first). |
| 8 | **`n:A:B` in WHERE vs MATCH**: `WHERE n:A` is a label predicate. This spec covers MATCH patterns; WHERE label predicates are a separate ticket. | Out of scope for this issue. |

---

### Scope Breakdown

| Phase | Work | Size |
|---|---|---|
| **Phase 1a** | Catalog side table + in-memory indexes + CREATE multi-label | M |
| **Phase 1b** | MATCH resolution (primary + secondary scan, intersection) | M |
| **Phase 1c** | `labels()` function update + `__labels__` metadata | S |
| **Phase 1d** | MERGE multi-label support | S |
| **Phase 2** | Cross-label property indexes | M |
| **Phase 3** | SET/REMOVE label mutations | L |

---

### Estimated Complexity

**Overall: L (Large)**

The parser is already done. The catalog and storage changes are moderate but the MATCH resolution touches ~15+ call sites in `scan.rs` and `hop.rs` that currently call `labels.first()`. Each must be updated to handle multi-label intersection semantics. Testing the combinatorics (single-label match on multi-label node, multi-label match, label-less scan including multi-label nodes, property index interaction) is substantial.

Recommended approach: Phase 1a+1b+1c+1d as a single PR (creates the feature end-to-end), Phase 2 and 3 as follow-ups.

---

### Appendix: Files to Modify

| File | Change |
|---|---|
| `crates/sparrowdb-catalog/src/catalog.rs` | Add label-set side table, reverse index, new TLV entry type |
| `crates/sparrowdb-catalog/src/tlv.rs` | New `NodeLabelSet` TLV tag (0x03) |
| `crates/sparrowdb-storage/src/wal/codec.rs` | New WAL op for label-set writes |
| `crates/sparrowdb-storage/src/wal/replay.rs` | Replay handler for label-set ops |
| `crates/sparrowdb-execution/src/engine/scan.rs` | Multi-label MATCH resolution (~15 call sites) |
| `crates/sparrowdb-execution/src/engine/hop.rs` | Multi-label hop traversal |
| `crates/sparrowdb-execution/src/engine/mutation.rs` | CREATE + MERGE multi-label support |
| `crates/sparrowdb-execution/src/engine/expr.rs` | `labels()` function update |
| `crates/sparrowdb-cypher/src/parser.rs` | MERGE multi-label parsing (minor) |

---

*Generated from codebase analysis of SparrowDB @ commit c7648d6.*

### LINKED ISSUE #415: Multi-label MATCH (n:B) on secondary label returns no rows — SPA-289 Phase 1 shipped broken
## Summary

5 of 10 tests in `crates/sparrowdb/tests/spa_289_multi_label.rs` fail on `main`. The multi-label feature landed in `de0b715` ("feat(catalog): multi-label nodes (n:A:B) — Phase 1 (closes #289)", #330) with its own acceptance tests failing.

Not caught because CI reported test failures as green — see #412. This is one of the failures that surfaced once the gate was fixed.

## Failing tests

```
match_on_secondary_label_finds_node
match_multi_label_pattern_returns_intersection
merge_single_label_then_match_multi_label_intersection
multi_label_match_discriminates_in_mixed_graph
three_label_node_all_patterns_match
```

## Reproduction

```rust
db.execute("CREATE (n:Animal:Pet {name: 'Rex'})").unwrap();
let result = db.execute("MATCH (n:Pet) RETURN n.name").unwrap();
// expected 1 row, got 0
```

```
assertion `left == right` failed: expected 1 row from secondary-label MATCH; got: []
  left: 0
 right: 1
```

Matching on the **primary** label works; matching on a **secondary** label finds nothing. Consistent with the secondary-label index not being written or not being consulted during label scan.

## Scope

- Reproduced on macOS (aarch64), clippy/rustc 1.97, against `main` at `fa536c0`.
- **Linux status unknown.** `cargo test` stops after a failing test binary, and `spa191_rel_type_persistence` fails earlier in the run, so `spa_289_multi_label` never executed on CI. It should be re-checked once #413 lands.
- Present in published **0.1.16** and **0.1.22** — the feature has been broken in released versions.

## Related

- #412 — CI pipefail fix that exposed this
- #413 — SPA-191 duplicate-edge fix, the other failure the same masking hid

ryaker:
Closes #415.

## Symptom

```
CREATE (n:Animal:Pet {name:'Rex'})
MATCH (n:Pet)         →  0 rows   (expected 1)
MATCH (n:Animal:Pet)  →  3 rows   (expected 1 — returned every Animal)
```

5 of 10 tests in `spa_289_multi_label.rs` have been failing since the feature landed in `de0b715` (#330, "closes #289") — the feature shipped with its own acceptance tests red, hidden by the CI masking fixed in #412. Present in published 0.1.16 and 0.1.22.

## Root cause — not the catalog

The secondary-label index was correct and populated all along. Two observations proved it:

- `labels(n)` returned `["Animal", "Pet"]` correctly
- the intersection worked in a single-node database

Only **execution-path selection** was wrong.

`execute_scan` (`scan.rs:1706`) routes to the chunked pipeline (#299) **before** the multi-label check (`scan.rs:1722`) and before the SPA-200 secondary-label union (`scan.rs:2110`). Both were therefore unreachable whenever the chunked path was eligible.

`can_use_chunked_pipeline` (`pipeline_exec.rs:81`) already keeps a list of fall-back-to-row-engine guards for features the chunked path doesn't implement — DISTINCT, inline prop filters, bare variable projection, `id(n)`. Multi-label was simply never added to it.

Two distinct defects followed:

| Defect | Effect |
|---|---|
| Chunked scan ignores every label after the first | `MATCH (n:Animal:Pet)` returned all Animals |
| Chunked scan walks only a label's own primary node store | `MATCH (n:Pet)` returned 0 rows |

## Fix

Add the two missing guards. The secondary-label check is a single HashMap lookup and only diverts queries the chunked path would otherwise answer incorrectly — queries on labels never used as secondary keep the fast path.

## Verification

- `spa_289_multi_label`: **10/10** (was 5/10)
- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace --exclude sparrowdb-ruby --all-targets -- -D warnings` clean
- Full `cargo test --workspace --exclude sparrowdb-ruby --no-fail-fast` — no regressions

## Pre-existing failures visible in that run (NOT caused by this PR)

Verified identical on unmodified `main`:

- `spa191_rel_type_persistence` — 2 failures, fixed by #413
- `vector_index::create_vector_index_ddl` — 1 failure: `CREATE VECTOR INDEX ... FOR ...` fails to parse with `InvalidArgument("expected LParen, got For")`
- `sparrowdb-bench ldbc_ic_test` — 10 failures, `IC*_stub_returns_*`

The last two need their own issues. They only became visible by running with `--no-fail-fast`: `cargo test` stops after a failing test binary, so `spa191` failing early hid everything downstream.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query execution compatibility by routing unsupported multi-label and unlabeled scans through the row-based engine.
  * Prevented incorrect use of the chunked scan path for queries involving secondary labels.

### DIFF
--- a/crates/sparrowdb-execution/src/engine/pipeline_exec.rs
+++ b/crates/sparrowdb-execution/src/engine/pipeline_exec.rs
@@ -120,33 +120,53 @@ impl Engine {
         if labels.is_empty() {
             return false;
         }
-        // SPA-289 (#415): the chunked scan is single-label and walks only the
-        // label's own primary node store. Two cases it answers incorrectly, both
-        // of which the row engine handles, so fall back to it:
-        //
-        // 1. Multi-label patterns. The chunked path ignores every label after
-        //    the first, so `MATCH (n:Animal:Pet)` returned all Animals instead
-        //    of the intersection. `execute_scan_multi_label` lives behind the
-        //    chunked check in execute_scan, so it was unreachable.
-        if labels.len() > 1 {
-            return false;
-        }
-        // 2. A label used as a *secondary* label. The chunked path can't see
-        //    those nodes at all — `CREATE (n:Animal:Pet)` then `MATCH (n:Pet)`
-        //    returned 0 rows. The row engine unions the primary scan with the
-        //    catalog's secondary-label reverse index.
-        //
-        // Cheap: one HashMap lookup, and it only diverts queries that would
-        // otherwise be wrong.
-        if let Ok(Some(label_id)) = self.snapshot.catalog.get_label(&labels[0]) {
-            if self
-                .snapshot
-                .catalog
-                .nodes_with_secondary_label(label_id)
-                .next()
-                .is_some()
-            {
-                return false;
+        // SPA-289 (#415): see all_chunked_node_labels_are_primary_only. Repeated
+        // here because can_use_chunked_pipeline is also called directly from
+        // execute_scan (scan.rs), not only via try_plan_chunked_match.
+        self.all_chunked_node_labels_are_primary_only(m)
+    }
+
+    /// Return `true` when every node in `m` can be resolved by primary label
+    /// alone — the only case the chunked executors handle correctly.
+    ///
+    /// Every chunked plan (Scan, OneHop, TwoHop, MutualNeighbors) resolves a
+    /// node pattern by taking `labels[0]` and walking that label's own primary
+    /// node store. Two shapes that breaks (SPA-289 / #415):
+    ///
+    /// 1. **Multi-label patterns** — labels after the first are ignored, so
+    ///    `MATCH (n:Animal:Pet)` matched every `Animal` instead of the
+    ///    intersection.
+    /// 2. **Labels used as a secondary label** — those nodes live in a
+    ///    different primary store and are invisible to the scan, so
+    ///    `CREATE (n:Animal:Pet)` then `MATCH (n:Pet)` returned 0 rows.
+    ///
+    /// The row-at-a-time engine handles both (intersection via
+    /// `execute_scan_multi_label`, secondary labels via the SPA-200 union), so
+    /// returning `false` here routes the query there.
+    ///
+    /// Cost: one catalog HashMap lookup per node, and it only diverts queries
+    /// the chunked path would otherwise answer incorrectly — a label never used
+    /// as a secondary label keeps the fast path.
+    fn all_chunked_node_labels_are_primary_only(&self, m: &MatchStatement) -> bool {
+        for pat in &m.pattern {
+            for node in &pat.nodes {
+                if node.labels.len() > 1 {
+                    return false;
+                }
+                let Some(label) = node.labels.first() else {
+                    continue;
+                };
+                if let Ok(Some(label_id)) = self.snapshot.catalog.get_label(label) {
+                    if self
+                        .snapshot
+                        .catalog
+                        .nodes_with_secondary_label(label_id)
+                        .next()
+                        .is_some()
+                    {
+                        return false;
+                    }
+                }
             }
         }
         true
@@ -680,6 +700,14 @@ impl Engine {
     /// MutualNeighbors is checked before TwoHop because it is a more specific
     /// pattern (both endpoints bound) that would otherwise fall into TwoHop.
     pub fn try_plan_chunked_match(&self, m: &MatchStatement) -> Option<ChunkedPlan> {
+        // SPA-289 (#415): every chunked plan resolves nodes by primary label
+        // only. Gate all of them, not just Scan — the one-hop, two-hop and
+        // mutual-neighbor executors take `labels[0]` and never consult the
+        // secondary-label index, so a relationship query would silently omit
+        // secondary-label nodes or ignore a multi-label intersection.
+        if !self.all_chunked_node_labels_are_primary_only(m) {
+            return None;
+        }
         // MutualNeighbors is a specialised 2-hop shape — check first.
         if self.can_use_mutual_neighbors_chunked(m) {
             return Some(ChunkedPlan::MutualNeighbors);
--- a/crates/sparrowdb-execution/src/engine/pipeline_exec.rs
+++ b/crates/sparrowdb-execution/src/engine/pipeline_exec.rs
@@ -116,7 +116,40 @@ impl Engine {
         if return_requires_row_engine(&m.return_clause.items) {
             return false;
         }
-        !m.pattern[0].nodes[0].labels.is_empty()
+        let labels = &m.pattern[0].nodes[0].labels;
+        if labels.is_empty() {
+            return false;
+        }
+        // SPA-289 (#415): the chunked scan is single-label and walks only the
+        // label's own primary node store. Two cases it answers incorrectly, both
+        // of which the row engine handles, so fall back to it:
+        //
+        // 1. Multi-label patterns. The chunked path ignores every label after
+        //    the first, so `MATCH (n:Animal:Pet)` returned all Animals instead
+        //    of the intersection. `execute_scan_multi_label` lives behind the
+        //    chunked check in execute_scan, so it was unreachable.
+        if labels.len() > 1 {
+            return false;
+        }
+        // 2. A label used as a *secondary* label. The chunked path can't see
+        //    those nodes at all — `CREATE (n:Animal:Pet)` then `MATCH (n:Pet)`
+        //    returned 0 rows. The row engine unions the primary scan with the
+        //    catalog's secondary-label reverse index.
+        //
+        // Cheap: one HashMap lookup, and it only diverts queries that would
+        // otherwise be wrong.
+        if let Ok(Some(label_id)) = self.snapshot.catalog.get_label(&labels[0]) {
+            if self
+                .snapshot
+                .catalog
+                .nodes_with_secondary_label(label_id)
+                .next()
+                .is_some()
+            {
+                return false;
+            }
+        }
+        true
     }
 
     /// Return `true` when `m` qualifies for Phase 2 one-hop chunked execution.
--- a/crates/sparrowdb/tests/spa_289_multi_label.rs
+++ b/crates/sparrowdb/tests/spa_289_multi_label.rs
@@ -445,3 +445,87 @@ fn multi_label_match_discriminates_in_mixed_graph() {
         sec_pet.rows[0][0]
     );
 }
+
+// ── Test 11: relationship patterns must also honour secondary labels ─────────
+
+/// Regression for the chunked one-hop/two-hop/mutual-neighbor plans.
+///
+/// `can_use_chunked_pipeline` guarded only the single-node Scan plan. The
+/// relationship plans in `try_plan_chunked_match` resolve each endpoint by
+/// `labels[0]` against that label's own primary store, so a hop whose endpoint
+/// is matched via a *secondary* label silently returned nothing.
+#[test]
+#[ignore = "SPA-289 Phase 1 is node-scan only: relationship patterns ignore \
+secondary labels in BOTH the chunked and row engines. Tracked in #419."]
+fn one_hop_match_honours_secondary_label_endpoint() {
+    let (_dir, db) = make_db();
+
+    // Rex is primary Animal, secondary Pet. Ann is a plain Person.
+    db.execute("CREATE (n:Animal:Pet {name: 'Rex'})")
+        .expect("CREATE Animal:Pet");
+    db.execute("CREATE (n:Person {name: 'Ann'})")
+        .expect("CREATE Person");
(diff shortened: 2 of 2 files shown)


ryaker:
Thanks — the finding was valid and is fixed in `a034996`.

Extracted `all_chunked_node_labels_are_primary_only` and applied it in `try_plan_chunked_match` before **any** plan is selected, covering MutualNeighbors, TwoHop, OneHop and Scan, across every node in every path pattern. `can_use_chunked_pipeline` calls the same helper, since `execute_scan` invokes it directly rather than through `try_plan_chunked_match`.

One correction to the proposed direction, found by writing the regression tests you asked for: **gating alone cannot fix relationship queries, because the row engine has the identical gap.** The SPA-200 secondary-label union and `execute_scan_multi_label` intersection both live only in the single-node scan path, so falling back routes to a different wrong answer rather than a right one:

| Query | Expected | Chunked | Row engine |
|---|---|---|---|
| `MATCH (a:Person)-[:OWNS]->(b:Pet)` | 1 | 0 | **0** |
| `MATCH (a:Person)-[:OWNS]->(b:Animal:Pet)` | 1 | 2 | **2** |

SPA-289 Phase 1 was node-scan only. The two regression tests land here `#[ignore]`d against #419, which tracks teaching the traversal path to resolve endpoints via `resolve_multi_label_node_ids`. Removing the `#[ignore]` is the acceptance criterion there.

The gating is still correct and worth keeping — it stops the chunked plans from being wrong in the cases the row engine gets right, and it makes the eventual #419 fix a single-engine change.

Issue #418, opened 2026-08-01
10 LDBC IC stub tests failing in sparrowdb-bench


ryaker:
## Summary

10 of 16 tests in `crates/sparrowdb-bench/tests/ldbc_ic_test.rs` fail on `main`:

```
ic3_stub_returns_empty      ic8_stub_returns_empty
ic4_stub_returns_empty      ic9_stub_returns_empty
ic5_stub_returns_empty      ic10_stub_returns_empty
ic6_stub_returns_empty      ic13_stub_returns_negative_one
ic7_stub_returns_empty      ic14_stub_returns_empty
```

Example: `IC3 stub must return empty Vec` (`ldbc_ic_test.rs:112`).

The tests assert that unimplemented LDBC Interactive Complex query stubs return empty results; they no longer do. Either the stubs gained behavior the tests weren't updated for, or they now error.

## Scope

- Reproduced on unmodified `main` at `fa536c0` (macOS aarch64, rustc 1.97).
- Confined to `sparrowdb-bench`, a benchmark crate — lower severity than #417, which affects a user-facing DDL path.
- Not caused by #413, #414, or #416 — verified identical on clean `main`.

## Why it was not caught

Same two layers as #417: CI masked failures via `tee` without `pipefail` (fixed in #412), and `cargo test` fail-fast meant this target never ran once an earlier binary failed. Only surfaced with `--no-fail-fast`.

## Suggestion

Worth deciding whether these stubs are still meaningful. If the IC queries are being implemented incrementally, the assertions should track implementation status rather than assert emptiness forever.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Aq3C5J646sBCH7sVs3uaCE

Issue #394, opened 2026-04-04
feat(vector): HNSW vector index + cosine/dot/euclidean similarity search


ryaker:
## Why

SparrowDB's primary use case is agent-native knowledge management (KMS). Without vector search, agents must use a separate vector store alongside SparrowDB, breaking the \"one database\" story. The goal: an agent should be able to write:

\`\`\`cypher
MATCH (n:Memory)
WHERE vector_similarity(n.embedding, \$query_vec) > 0.8
RETURN n ORDER BY vector_similarity(n.embedding, \$query_vec) DESC LIMIT 10
\`\`\`

...and get results from a single embedded store with no external dependencies.

### Value type
- Add \`Value::Vector(Vec<f32>)\` to \`sparrowdb-execution\`
- PackStream encoding for Bolt: serialize as typed list

### Storage
- New \`VectorIndex\` struct in \`sparrowdb-storage\`
- Per-(label, property) HNSW index stored in \`<db>/vector_indexes/<label>_<prop>.hnsw\`
- Persist index to disk (bincode or custom format); rebuild from WAL on open if missing
- HNSW parameters: \`M\` (max connections, default 16), \`ef_construction\` (default 200)

### Index management (DDL)
\`\`\`cypher
CREATE VECTOR INDEX FOR (n:Memory) ON (n.embedding) OPTIONS { dimensions: 1536, similarity: 'cosine' }
DROP INDEX memory_embedding_vector
\`\`\`

### Query functions (Cypher)
| Function | Returns | Notes |
|----------|---------|-------|
| \`vector_similarity(prop, \$vec)\` | Float64 0..1 | cosine by default |
| \`vector_distance(prop, \$vec)\` | Float64 | euclidean |
| \`vector_dot(prop, \$vec)\` | Float64 | dot product |
| \`vector_search(label, prop, \$vec, k)\` | List<NodeRef> | ANN top-K, no full MATCH needed |

### Planner
- Detect \`vector_similarity(...) > threshold\` or \`ORDER BY vector_similarity(...) LIMIT k\` in WHERE/ORDER clauses
- Route to HNSW ANN scan instead of full table scan when index exists

### Parameters
- \`$query_vec\` passed as \`List<Float64>\` in Bolt RUN params map

## Implementation Notes

- Use pure-Rust HNSW: \`hnsw_rs\` crate (no C deps) or implement directly (~400 lines)
- Index lives alongside CSR files; no new storage backend needed
[...]

PR #420, opened 2026-08-01
fix(cypher): CREATE VECTOR INDEX could not be parsed (closes #417)

### LINKED ISSUE #417: CREATE VECTOR INDEX fails to parse: expected LParen, got For
## Summary

`crates/sparrowdb/tests/vector_index.rs::create_vector_index_ddl` fails on `main`:

```
CREATE VECTOR INDEX must succeed: InvalidArgument("expected LParen, got For")
  crates/sparrowdb/tests/vector_index.rs:29
```

The parser rejects the `FOR` keyword in the vector-index DDL, so a vector index cannot be created through Cypher at all. The HNSW index feature (#394 / #398) is advertised but its DDL entry point does not parse.

## Scope

- Reproduced on unmodified `main` at `fa536c0` (macOS aarch64, rustc 1.97).
- Present in published **0.1.22**.
- Not caused by #413, #414, or #416 — verified identical on clean `main`.

## Why it was not caught

Two layers:

1. CI reported test failures as green — `cargo test … | tee` with no `pipefail` (fixed in #412).
2. `cargo test` halts after a failing test binary, so `spa191_rel_type_persistence` failing earlier meant this target never ran. Only surfaced with `--no-fail-fast`.

## Related

- #412 — CI gating fix
- #415 / #416 — multi-label MATCH, same masking
- Companion issue for the 10 failing `ldbc_ic_test` stubs in `sparrowdb-bench`

ryaker:
Closes #417.

## Symptom

```cypher
CREATE VECTOR INDEX FOR (n:Memory) ON (n.embedding)
OPTIONS { dimensions: 4, similarity: 'cosine' }
```

```
InvalidArgument("expected LParen, got For")
```

The HNSW vector index (#394) could not be created through Cypher **at all**. Shipped broken in published 0.1.16 and 0.1.22.

## Root cause

The lexer maps the `FOR` keyword to a dedicated `Token::For` (`lexer.rs:393`) and never to `Token::Ident("FOR")`.

`parse_create_vector_index` matched only the ident form:

```rust
match self.peek().clone() {
    Token::Ident(ref s) if s.eq_ignore_ascii_case("FOR") => { self.advance(); }
    _ => {} // tolerate missing FOR
}
self.expect_tok(&Token::LParen)?;   // ← sees Token::For, errors
```

`Token::For` fell through the tolerate-missing-FOR arm without advancing, so the very next `expect_tok(&Token::LParen)` failed on it.

`parse_create_fulltext_index` gets this right — it uses `Token::For` at `parser.rs:1447` and `1456`. The vector-index parser, written later, copied the shape but not the token.

## Fix

Add the `Token::For` arm. The ident arm and the tolerate-missing-`FOR` fallback are both kept, so nothing previously accepted regresses.

## Verification

- `vector_index` integration tests: **16/16** (was 15/16)
- `spa_fulltext`: 5/5 — `CREATE FULLTEXT INDEX` unaffected
- `cargo test -p sparrowdb-cypher` — all green
- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace --exclude sparrowdb-ruby --all-targets -- -D warnings` clean

## Context

Fourth bug found while verifying 0.1.22 for release, all of which were invisible because CI reported test failures as green (#412) and `cargo test` fail-fast hid everything after the first failing binary. With this and #416 merged, the only remaining known failure on `main` is #418 (LDBC stubs in `sparrowdb-bench`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of vector index creation statements using the `FOR` keyword.
  * Preserved compatibility with identifier-form and omitted `FOR` syntax.

### DIFF
--- a/crates/sparrowdb-cypher/src/parser.rs
+++ b/crates/sparrowdb-cypher/src/parser.rs
@@ -1498,7 +1498,16 @@ impl Parser {
         };
 
         // `FOR (n:Label)`
+        //
+        // The lexer maps the FOR keyword to the dedicated `Token::For`
+        // (lexer.rs), never to `Token::Ident("FOR")`, so matching only on the
+        // ident form left the token unconsumed and the `expect_tok(LParen)`
+        // below failed with "expected LParen, got For" — CREATE VECTOR INDEX
+        // could not be parsed at all (#417).
         match self.peek().clone() {
+            Token::For => {
+                self.advance();
+            }
             Token::Ident(ref s) if s.eq_ignore_ascii_case("FOR") => {
                 self.advance();
             }


ryaker:
@coderabbitai review

Issue #421, opened 2026-08-01
Variable-length path followed by another hop silently truncates to 1 hop (wrong results, no error)


ryaker:
## Summary

When a variable-length relationship pattern is followed by a further hop, the variable-length expansion collapses to **one hop**. Matches at depth ≥ 2 are silently dropped. No error is raised — the query returns a plausible but incomplete result set.

## Isolated reproduction

Using `crates/sparrowdb-bench/fixtures/ldbc/mini`, Alice is `ldbc_id: 1`.

```cypher
-- A: variable-length alone — CORRECT (8 rows, includes Dave/Eve/Grace at depth 2)
MATCH (p:Person {ldbc_id: 1})-[:knows*1..2]->(f:Person)
RETURN f.firstName

-- B: same pattern + one more hop — WRONG (3 rows: Bob, Carol, Frank)
MATCH (p:Person {ldbc_id: 1})-[:knows*1..2]->(f:Person)-[:isLocatedIn]->(pl:Place)
RETURN f.firstName, pl.name
```

B returns exactly the **1-hop** neighbour set `{Bob, Carol, Frank}`. The depth-2 matches are gone.

They are not missing for lack of data — every person has a location:

```cypher
MATCH (f:Person)-[:isLocatedIn]->(pl:Place) RETURN f.firstName, pl.name
-- 10 rows, including Dave → Germany
```

Fixture edges (`person_knows_person_0_0.csv`): `1→2, 1→3, 1→6` at depth 1; `2→4, 3→4, 3→5, 2→5, 6→7` giving `{4,5,7}` at depth 2. Dave (person 4) is in Germany (place 3) and must appear in B.

## Impact

Any `MATCH (a)-[:R*m..n]->(b)-[:S]->(c)` shape returns under-complete results. This is a **silent correctness bug** — the worst kind, since callers cannot detect it.

It is the underlying cause of #418: `ic3_friends_in_countries` uses exactly this shape and returns only `[("Frank","Miller")]` when the correct answer for Alice/Germany is `[Dave Brown, Frank Miller]`.

## Scope

- Reproduced on `main` at `25f3559` (macOS aarch64, rustc 1.97).
- Present in published 0.1.16 and 0.1.22.
- Not caused by any 2026-07-31/08-01 PR.

## Relationship to #418

[...]

PR #425, opened 2026-08-01
fix(bench): ic6 applies its tag filter; ic13 returns -1 when unreachable (closes #422, #423)

### LINKED ISSUE #422: ic6_tag_co_occurrence ignores the input tag filter (returns co-occurring tags for tags that do not exist)
## Summary

`ic6_tag_co_occurrence(db, person, tag)` does not restrict to posts carrying `tag`. It only removes `tag` from the output, so any tag — including one absent from the graph — returns the full set of friends' tags.

## Reproduction

`crates/sparrowdb-bench/fixtures/ldbc/mini`, person 1 (Alice), friends `{2,3,6}`:

| Call | Returns | Correct |
|---|---|---|
| `ic6(1, "Rust")` | `[("SocialNetworks", 1)]` | ❌ expected empty — see correction below |
| `ic6(1, "SocialNetworks")` | `[("Rust", 1)]` | ❌ expected empty — see correction below |
| `ic6(1, "Databases")` | `[("Rust",1), ("SocialNetworks",1)]` | ❌ expected empty |
| `ic6(1, "Music")` | `[("Rust",1), ("SocialNetworks",1)]` | ❌ expected empty — no such tag |
| `ic6(1, "ZZZ_nonexistent")` | `[("Rust",1), ("SocialNetworks",1)]` | ❌ expected empty |

## Correction (2026-08-01)

The first two rows were originally marked correct. **They are also wrong**, caught in review. Friend posts are post 2 (Bob, tagged Rust) and post 3 (Carol, tagged SocialNetworks). `ic6(1,"Rust")` filters to friend posts tagged Rust — post 2 alone — which carries no other tag, so the correct answer is **empty**. SocialNetworks sits on post 3, which is not tagged Rust, so it must not co-occur. Symmetrically for `ic6(1,"SocialNetworks")`.

Every row in the table above is therefore wrong. The first two merely *looked* right because, with one tag per friend post, "exclude the input tag" coincides with the correct answer. A neat illustration of why an assertion must be derived, not observed.

## Why those are wrong

LDBC IC6 asks: among posts **tagged with the given tag** created by the person's friends, which other tags co-occur?

- `Databases` (tag 1) appears only on posts 1 and 4, both created by Alice herself — no *friend* post carries it, so the result must be empty.
- `Music` and `ZZZ_nonexistent` are not in `tag_0_0.csv` at all, so nothing can match.

Returning results in these cases means the "posts tagged with X" restriction is never applied. The two passing cases pass only because, with two friend posts carrying one tag each, "exclude X" happens to coincide with the right answer.

## Scope

- Reproduced on `main` at `25f3559` (macOS aarch64).
- Confined to `sparrowdb-bench`; no impact on the core engine.
- Distinct from #421 — `ic6` uses no variable-length path.

## Relationship to #418

`ic6_stub_returns_empty` in `ldbc_ic_test.rs` is one of the stale stub-era assertions. It cannot be given a correct non-empty assertion until this is fixed; a derived assertion here would encode the bug.

### LINKED ISSUE #423: ic13_shortest_path returns 1 instead of -1 for unreachable pairs
## Summary

`ic13_shortest_path` documents:

> Returns the hop count, or **-1 if no path exists**.

For an unreachable pair it returns `1`.

## Reproduction

`crates/sparrowdb-bench/fixtures/ldbc/mini`:

```rust
ic_queries::ic13_shortest_path(&db, 10, 1)   // → 1, expected -1
ic_queries::ic13_shortest_path(&db, 1, 10)   // → 5, correct
```

`knows` edges are directed. Person 10 appears only as a **target** (`9|10` in `person_knows_person_0_0.csv`) and has no outgoing edges, so no path 10→1 exists. The forward direction is correct: `1→6→7→8→9→10` is 5 hops, shorter than `1→3→5→7→8→9→10` (6) or `1→2→4→6→7→8→9→10` (7).

Returning `1` is additionally suspicious because it is not a plausible path length between these two nodes in any direction — it looks like a sentinel or count leaking through rather than a computed distance.

## Impact

Callers cannot distinguish "adjacent" from "unreachable". Any code branching on `-1` silently treats disconnected pairs as directly connected.

## Scope

- Reproduced on `main` at `25f3559` (macOS aarch64).
- Not attributed yet between `ic13_shortest_path`'s extraction of the `shortestPath(...)` result and the engine's `shortestPath` implementation — worth checking the raw Cypher `MATCH (a:Person {ldbc_id: 10}), (b:Person {ldbc_id: 1}) RETURN shortestPath((a)-[:knows*]->(b))` before deciding where to fix.
- Independent of #421 (no variable-length-plus-hop shape) and #422.

## Test

`ic13_unreachable_returns_negative_one` in `crates/sparrowdb-bench/tests/ldbc_ic_test.rs` is `#[ignore]`d against this issue. Un-ignoring it is the acceptance criterion.

ryaker:
Closes #422. Closes #423.

**Stacked on #424** (`fix/varlen-truncation-guard`, rebased onto head `2a6890c`) — that PR introduces the two `#[ignore]`d tests this one un-ignores. Retarget to `main` once #424 lands.

Related: **#427** (the engine bug behind #423, filed separately — not fixed here). **#428** (the fixture gap behind IC6's liveness guard, filed from this work).

## #422 — ic6 ignored its input tag filter

`ic6_tag_co_occurrence` never restricted to posts carrying the given tag. It collected every post created by the person's friends, counted every tag on them, and merely excluded the input tag from the output. `ic6(1, "Databases")` returned `[(Rust,1), (SocialNetworks,1)]` even though Databases sits only on Alice's own posts 1 and 4; `ic6(1, "ZZZ_nonexistent")` returned the same full set.

Fix: added the missing step — narrow the friends' posts to those actually carrying the tag, then count the other tags on exactly those posts.

**Issue #422's table is wrong on one row.** It marks `ic6(1, "Rust") → [(SocialNetworks,1)]` as correct. It is not. Post 2 is the only friend post tagged Rust and carries no other tag, so the correct LDBC answer is empty; SocialNetworks lives on post 3, which is not tagged Rust. Both "looks right" rows in #422 passed by coincidence — with one tag per friend post, "exclude the input tag" happens to coincide with the right answer.

## #423 — ic13 returned 1 instead of -1 — **root cause is in the engine**

Per the investigation instruction, I ran the raw Cypher before deciding where to fix:

```
MATCH (a:Person {ldbc_id: 10}), (b:Person {ldbc_id: 1}) RETURN shortestPath((a)-[:knows*]->(b))
  → Int64(1)      ← engine, wrong (no path exists)
MATCH (a:Person {ldbc_id: 2}),  (b:Person {ldbc_id: 1}) RETURN shortestPath((a)-[:knows*]->(b))
  → Int64(1)      ← engine, wrong (1 is never a knows target)
```

The engine is at fault, in two independent ways, both in `crates/sparrowdb-execution`:

1. **The relationship-type constraint is dropped.** `eval_shortest_path_expr` (`engine/expr.rs:526`) calls `bfs_shortest_path`, which calls `get_node_neighbors_by_slot(..., &[])` (`engine/path.rs:585`). The empty `rel_ids` slice means "no type filter", so the BFS for `[:knows*]` walks `isLocatedIn`, `likes`, `hasCreator` and every other edge type.
2. **Arrival is decided by slot, ignoring label.** `bfs_shortest_path` tests `nb_slot == dst_slot` (`engine/expr.rs:587`). Slot 0 of Place, Post, and Person are all `0`, so the slot-N node of *any* label terminates the BFS.

On the mini fixture the two combine exactly: person 10's only outgoing edge is `isLocatedIn` → Place id 1 → Place **slot 0**, and target person 1 is Person **slot 0**. Hence the reported 1-hop path between two unconnected persons. Same mechanism for `2→1`: person 2 `likes` post 1 → Post slot 0.

Engine files are outside this PR's scope, so **the engine bug is not fixed here** — it is filed as **#427**, which also notes that `visited.insert(nb_slot)` has the same slot/label conflation and can prune legitimate paths. What is fixed is the bench layer: IC13 now does its own level-synchronous BFS over `knows`, one Cypher query per hop, keyed on `ldbc_id` rather than storage slots. This follows the workaround pattern the module already documents for other engine limitations.

## Also — ic4/ic6 ordering determinism

CodeRabbit flagged on #424 that `ic4_top_tags` ordered by `cnt DESC` only, so tied counts could come back in any order and an exact-vector assertion was flaky. Added `tag.name ASC` as a deterministic tie-break in both IC4 and IC6.

## Tests

`ic6_co_occurring_tags` and `ic13_unreachable_returns_negative_one` un-ignored. Every expectation derived by hand from the fixture CSVs, none captured from program output.

Both were verified to **fail** against the unfixed `ic_queries.rs` with exactly the symptoms in the issues (`got 1`; `got [("SocialNetworks", 1)]`).

Assertions added beyond the un-ignore:

- **IC6**: `SocialNetworks` (symmetric single-tag case) and `Databases` (#422's headline case).
- **IC6 liveness guard.** Every IC6 expectation the mini fixture can express is empty: the only multi-tag post is 1 (`{Databases, GraphTheory}`), it belongs to Alice, and nobody `knows` Alice — person 1 is never a `knows` target — so post 1 is never anyone's friend post. That makes all four assertions satisfiable by an `ic6` that unconditionally returns empty. The test therefore also asserts IC4 (same friend→post→tag pipeline, minus the tag restriction) is non-empty, so the empties are provably the restriction at work rather than a dead query. Full derivation and a proposed fixture change are in **#428**.
- **IC13**: `2→1` unreachable (source *has* outgoing edges but 1 is never a `knows` target — separates "dead end" from "genuinely unreachable", and this is the case the pre-fix engine also got wrong), and absent persons in both directions.
- **New `ic13_directed_and_self_paths`**: `1→2 = 1`, `1→1 = 0`, `9999→9999 = -1`, `1→9 = 4` (`1→6→7→8→9`, vs 5 for `1→3→5→7→8→9` and 6 for `1→2→4→6→7→8→9`).

## Sibling test corrected

`crates/sparrowdb/tests/spa_111_ldbc_ic3_ic14.rs::ic6_tag_co_occurrence` asserted `ic6(1, "Rust")` is **non-empty**, under the comment *"Exclude 'Rust' — should find other tags on friend posts"* — a second copy of the #422 bug written down as the expectation. Corrected to assert empty, plus the unknown-tag case and the same liveness guard. Taken into this PR at the coordinator's request so the corrected assertion and the fix that makes it pass land together.

The file's module doc claimed *"Each test asserts non-empty results"*; updated, since non-empty was never the goal and treating it as one is what let #422 hide.

## Verification

Run with an isolated `CARGO_TARGET_DIR` — worktrees in this repo share one by default, which can silently produce a test binary built from another worktree's source.

```
CARGO_TARGET_DIR=/Volumes/Dev/target-425 cargo test -p sparrowdb-bench -p sparrowdb --no-fail-fast
CARGO_TARGET_DIR=/Volumes/Dev/target-425 cargo fmt --all -- --check
CARGO_TARGET_DIR=/Volumes/Dev/target-425 cargo clippy -p sparrowdb-bench --all-targets -- -D warnings
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tag-count query results with deterministic alphabetical ordering for ties.
  * Corrected tag co-occurrence results for friend posts and unknown tags.
  * Improved shortest-path queries with directed traversal, bounded search, and proper handling of missing or unreachable endpoints.

* **Tests**
  * Re-enabled integration coverage for previously skipped queries.
  * Added validation for empty results, self-paths, missing people, unreachable paths, and pipeline behavior.

### DIFF
--- a/crates/sparrowdb-bench/src/ic_queries.rs
+++ b/crates/sparrowdb-bench/src/ic_queries.rs
@@ -14,7 +14,7 @@
 
 use sparrowdb::GraphDb;
 use sparrowdb_execution::Value;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 // ── Result types ──────────────────────────────────────────────────────────────
 
@@ -138,6 +138,10 @@ pub fn ic3_friends_in_countries(
 /// **IC4** — Tags on posts created by direct friends.
 ///
 /// Two-step: first get friend IDs, then query posts by those creators.
+///
+/// Ties on `cnt` are broken by `tag.name ASC` so the ordering is total and the
+/// result is reproducible across runs — otherwise equal-count tags come back in
+/// whatever order the aggregation happens to produce.
 pub fn ic4_top_tags(
     db: &GraphDb,
     person_id: i64,
@@ -174,7 +178,7 @@ pub fn ic4_top_tags(
         "MATCH (post:Post)-[:hasTag]->(tag:Tag) \
          WHERE post.ldbc_id IN [{}] \
          RETURN tag.name, COUNT(*) AS cnt \
-         ORDER BY cnt DESC LIMIT 10",
+         ORDER BY cnt DESC, tag.name ASC LIMIT 10",
         post_ids_list.join(", ")
     );
     let result = db.execute(&tag_query)?;
@@ -218,9 +222,23 @@ pub fn ic5_forums_with_friends(
 
 // ── IC6 — Tag co-occurrence ─────────────────────────────────────────────────
 
-/// **IC6** — Tags on posts by friends, excluding a given tag.
+/// **IC6** — Tags that co-occur with `tag_name` on the person's friends' posts.
 ///
-/// Two-step: reuses IC4 post discovery, filters tags.
+/// LDBC IC6 asks: among the posts *carrying* `tag_name` that were created by the
+/// person's friends, which other tags appear? Three steps, because the engine
+/// cannot chain these hops in one pattern:
+///
+/// 1. friends of `person_id`,
+/// 2. posts created by those friends,
+/// 3. of those, the posts carrying `tag_name` — then count the *other* tags on
+///    exactly those posts.
+///
+/// Step 3's first half is the restriction that was missing (#422): the query
+/// previously counted every tag on every friend post and merely excluded
+/// `tag_name` from the output, so a tag carried by no friend post — or absent
+/// from the graph entirely — still returned the full set of friends' tags.
+///
+/// Ties on `cnt` are broken by `tag.name ASC` for a reproducible ordering.
 pub fn ic6_tag_co_occurrence(
     db: &GraphDb,
     person_id: i64,
@@ -249,14 +267,36 @@ pub fn ic6_tag_co_occurrence(
         return Ok(Vec::new());
     }
 
+    // Step 2: narrow the friends' posts to those actually carrying `tag_name`.
+    // Without this the tag argument only ever suppressed a row in the output.
     let post_ids_list: Vec<String> = post_ids.iter().map(|id| id.to_string()).collect();
+    let tagged_query = format!(
+        "MATCH (post:Post)-[:hasTag]->(tag:Tag) \
+         WHERE post.ldbc_id IN [{}] AND tag.name = $tagName \
+         RETURN post.ldbc_id",
+        post_ids_list.join(", ")
+    );
+    let tagged_params = HashMap::from([("tagName".into(), Value::String(tag_name.into()))]);
+    let tagged_result = db.execute_with_params(&tagged_query, tagged_params)?;
+    let tagged_post_ids: Vec<i64> = tagged_result
+        .rows
+        .iter()
+        .map(|row| value_to_i64(&row[0]))
+        .collect();
+
+    if tagged_post_ids.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    // Step 3: count the other tags carried by exactly those posts.
+    let tagged_ids_list: Vec<String> = tagged_post_ids.iter().map(|id| id.to_string()).collect();
     let params = HashMap::from([("tagName".into(), Value::String(tag_name.into()))]);
     let tag_query = format!(
         "MATCH (post:Post)-[:hasTag]->(tag:Tag) \
          WHERE post.ldbc_id IN [{}] AND tag.name <> $tagName \
          RETURN tag.name, COUNT(*) AS cnt \
-         ORDER BY cnt DESC LIMIT 10",
-        post_ids_list.join(", ")
+         ORDER BY cnt DESC, tag.name ASC LIMIT 10",
+        tagged_ids_list.join(", ")
     );
     let result = db.execute_with_params(&tag_query, params)?;
     Ok(result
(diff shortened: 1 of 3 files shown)


ryaker:
@coderabbitai review

The earlier attempt hit the account PR review limit, so the check went green without a review having happened. Requesting a real one.

ryaker:
@coderabbitai full review

Rebased onto `6254f91`. The previous CodeRabbit run on this PR returned `Review rate limited` — no actual review happened, so this is the first real pass.

Changes since your last look:
- rebased off the pre-#412 CI, so this now gets graded by the pipefail gate (the old green was a 2m11s run that aborted at the first failing binary with the exit code swallowed)
- un-ignored `ic3_friends_in_countries_returns_both_germans` and `ic11_job_referral`, which were skipped as "blocked on #421"; #421 shipped in the rebase base

Verified falsifiable against the pre-fix source (`ic_queries.rs` reverted to `6254f91`, tests held at HEAD): `ic6` fails with `got [("SocialNetworks", 1)]` and `ic13_unreachable_returns_negative_one` fails with `expected -1; got 0`. Both are the reported symptoms of #422/#423.

Please pay particular attention to whether `ic6_co_occurring_tags` proves anything: every IC6 expectation the mini fixture can express is empty (see #428), so the assertions are guarded by an IC4 liveness check rather than a positive result.

ryaker:
@coderabbitai full review

Retrying: the previous request returned `Action not completed — Review rate limited` inside a collapsed details block, so no review has run on this PR yet.

ryaker:
@coderabbitai full review

Third attempt — the previous two both returned `Review rate limited`, so no review has run on this PR. Requesting alone this time so it does not contend with another PR for the included-review quota.

Issue #431, opened 2026-08-02
get_node_neighbors_labeled falls back to the source node's label, so cross-label variable-length traversal breaks after checkpoint


ryaker:
## Summary

`Engine::get_node_neighbors_labeled` (`crates/sparrowdb-execution/src/engine/path.rs`) recovers a CSR neighbour's label by guessing from delta-log hints, and falls back to the **source** node's label when no hint exists:

```rust
if !found {
    // No label info available — fallback to src_label_id (correct for
    // homogeneous graphs, gracefully wrong for unmapped CSR-only nodes
    // in heterogeneous graphs with no delta activity on those nodes).
    out.insert((dst_slot, src_label_id));
}
```

The `node_label` hint set and `all_label_ids` are both built from `read_delta_all()`. After `checkpoint()` the delta log is empty, so **every** hint disappears and **every** CSR neighbour gets the source node's label — regardless of what it actually is.

The comment calls this "gracefully wrong". It is not graceful: it is silent and it fires on every checkpointed database, which is every database that has been open for any length of time.

## Symptom

A variable-length traversal that crosses labels returns correct results before `checkpoint()` and wrong (usually empty) results after.

Two independent confirmations on `crates/sparrowdb-bench/fixtures/ldbc/mini`:

1. `shortestPath((p:Person {ldbc_id: 1})-[:isLocatedIn*]->(pl:Place))` — correct uncheckpointed, NULL after checkpoint. Found while fixing #427.
2. Variable-length MATCH with a destination label that differs from the source label. `execute_variable_length` (path.rs) and, since #421, the varlen hop inside `execute_n_hop` both filter destinations with:

   ```rust
   if let Some(required_label) = dst_label_id {
       if actual_label_id != required_label { continue; }
   }
   ```

   With every `actual_label_id` collapsed to the source label, that filter drops everything. `MATCH (p:Person)-[:R*1..2]->(x:Post)` silently returns zero rows on a checkpointed graph.

## Why it survives testing

Homogeneous traversals — `knows*1..2` over Person→Person, which is what most varlen tests use — are unaffected, because the source label *is* the correct destination label. The bug only appears when a varlen hop crosses labels, and only after checkpoint. LDBC IC3/IC11 do not catch it because their varlen segment is Person→Person; only the trailing fixed hops cross labels, and those go through a different code path.

## Suggested fix

[...]

Issue #430, opened 2026-08-01
MATCH after WITH silently drops all hops but the first (execute_pipeline_match_stage)


ryaker:
## Summary

A MATCH stage that follows a `WITH` silently drops every relationship after the first. No error; the query returns rows with `NULL` for the variables it never bound, or — worse — rows that look like a complete answer to a shorter pattern than the one that was written.

## Where

`crates/sparrowdb-execution/src/engine/scan.rs`:

```rust
// execute_pipeline_match_stage
if !pat.rels.is_empty() {
    // Relationship traversal in a pipeline MATCH stage.
    // Currently supports single-hop: (src)-[:REL]->(dst)
    return self.execute_pipeline_match_hop(pat, where_clause, binding);
}
```

`execute_pipeline_match_hop` then reads only:

```rust
let src_pat = &pat.nodes[0];
let dst_pat = &pat.nodes[1];
let rel_pat = &pat.rels[0];
```

`pat.rels[1..]` and `pat.nodes[2..]` are never looked at. The comment says "currently supports single-hop", but nothing rejects the patterns it does not support — it just answers a different query.

## Reproduction shape

```cypher
MATCH (a:Person {name: 'Alice'})
WITH a
MATCH (a)-[:knows]->(b:Person)-[:livesIn]->(c:City)
RETURN b.name, c.name
```

`c` is never bound. The stage returns one binding per `b`, and `c.name` projects as `NULL` — but the row count is that of the one-hop query, so the result reads as "Alice's friends have no city" rather than "this pattern was not executed".

## Relationship to #421

#421 was the same failure in a different executor: a pattern shape that the dispatcher did not recognise got answered by a narrower executor instead of being rejected. The #421 PR adds an explicit `InvalidArgument` in `execute_pipeline_match_hop` for the *variable-length* case (`WITH a MATCH (a)-[:R*1..2]->(b)`), because a quantifier there was being truncated to depth 1 exactly as in #421.

The multi-hop case is left as-is by that PR: it is pre-existing, not varlen-specific, and rejecting it might break callers that are (unknowingly) relying on the truncated behaviour. Hence this issue.

## Suggested fix

Two options, in order of preference:

1. **Implement it.** `execute_pipeline_match_hop` could iterate `pat.rels` the way `execute_n_hop` does, threading the accumulated binding map forward. The binding-map representation is already compatible.
[...]

Issue #429, opened 2026-08-01
execute_n_hop unions CSRs across rel tables of the same type, so nodes inherit another label's edges (silent wrong rows)


ryaker:
## Summary

`execute_n_hop` (the 3+-hop chain traversal in `crates/sparrowdb-execution/src/engine/hop.rs`) resolves relationship-table IDs by **type name alone**:

```rust
let rel_ids_per_hop: Vec<Vec<u32>> = (0..n_rels)
    .map(|i| self.resolve_rel_ids_for_type(&pat.rels[i].rel_type))
    .collect();
```

`resolve_rel_ids_for_type` returns **every** rel table whose type name matches, ignoring the endpoint labels. Those IDs are then unioned in `csr_neighbors_filtered(src_slot, rel_ids)` — and a `CsrForward` is indexed by **source slot**, with slots numbered per label.

So when one relationship type backs more than one `(src_label, dst_label)` pair, a node of label A inherits the edges of the node of label B that happens to occupy the same slot number.

This is a **silent correctness bug**: no error, plausible-looking rows, extra results the caller cannot detect.

## Reproduction

`crates/sparrowdb-bench/fixtures/ldbc/mini` registers `isLocatedIn` twice:

- `(Person)-[:isLocatedIn]->(Place)`
- `(Organisation)-[:isLocatedIn]->(Place)`

```cypher
MATCH (p:Person {ldbc_id: 1})-[:knows]->(f:Person)-[:knows]->(g:Person)
      -[:isLocatedIn]->(place:Place)
WHERE place.name = 'Germany'
RETURN DISTINCT g.firstName, g.lastName
```

Carol White comes back. Carol is `ldbc_id` 3 → **Person slot 2**. `person_isLocatedIn_place_0_0.csv` puts her in place 1 (United States). But **Organisation slot 2** is org 3, TechStart Inc, and `organisation_isLocatedIn_place_0_0.csv` puts org 3 in place 3 (Germany). The two CSRs are unioned, so Carol picks up TechStart's location.

This was found while implementing #421 — LDBC IC3 for `(person 1, "Germany")` returned `[Dave Brown, Frank Miller, Carol White]` when the correct answer is `[Dave Brown, Frank Miller]`.

## Why one- and two-hop queries are unaffected

[...]

Issue #427, opened 2026-08-01
shortestPath() ignores relationship-type filter and matches destination by slot without label — wrong results on any multi-label or multi-edge-type graph


ryaker:
## Severity

`shortestPath()` returns **wrong distances on essentially any real graph** — anything with more than one node label or more than one relationship type. Silent: no error, a plausible small integer. Present in published 0.1.16 and 0.1.22.

## Reproduction

LDBC mini fixture (`crates/sparrowdb-bench/fixtures/ldbc/mini`), independently verified:

```cypher
MATCH (x:Person {ldbc_id: 10}), (y:Person {ldbc_id: 1})
RETURN shortestPath((x)-[:knows*]->(y))
```

| Query | Returns | Correct |
|---|---|---|
| `sp(10 → 1)` | **1** | no path — person 10 has no outgoing `knows` |
| `sp(2 → 1)` | **1** | no path — person 2 knows only 4 and 5 |
| `sp(1 → 10)` | 5 | ✅ |
| `sp(1 → 2)` | 1 | ✅ |

## Root cause — two independent faults in `crates/sparrowdb-execution/src/engine/expr.rs` (`bfs_shortest_path`)

**1. The relationship-type filter is discarded.**

```rust
let neighbors =
    self.get_node_neighbors_by_slot(node_slot, node_label_id, &delta_idx, &[]);
```

The empty `rel_ids` slice means "no type filter", so a BFS for `[:knows*]` traverses `isLocatedIn`, `likes`, `hasCreator` — every edge type in the graph.

**2. The destination test compares storage slots without comparing labels.**

```rust
for nb_slot in neighbors {
    if nb_slot == dst_slot {
        return Some(depth);
    }
    if visited.insert(nb_slot) { ... }
```

`nb_slot` is a per-label storage slot. Slot 0 of `Person`, `Place` and `Post` are all `0`, so a neighbour in a completely different label counts as reaching the target. `visited` has the same conflation — keyed on slot alone, so nodes in different labels alias each other and legitimate paths can be pruned.

**They combine exactly.** Person 10's only outgoing edge is `isLocatedIn` → Place slot 0; target Person 1 is Person slot 0 → reports distance 1. Person 2 `likes` post 1 → Post slot 0 → same.

## Impact

- Any `shortestPath()` result is untrustworthy unless the graph has exactly one label and one edge type.
- Wrong in the *unsafe* direction: reports paths that do not exist, and understates real distances.
[...]

PR #433, opened 2026-08-02
feat(engine): variable-length path followed by further hops (closes #421)

### LINKED ISSUE #421: Variable-length path followed by another hop silently truncates to 1 hop (wrong results, no error)

ryaker:
Closes #421. Replaces the interim guard from #424 with real support.

## The bug

```cypher
MATCH (a)-[:R*1..2]->(b)-[:S]->(c)   -- returned only depth-1 matches, no error
```

`execute_match` dispatched on relationship count. `is_var_len` required the
varlen relationship to be the **only** relationship in the pattern, so a
trailing hop made `rels.len() == 2`, `is_two_hop` won, and `execute_two_hop`
treated `*1..2` as a plain single hop — silently discarding every match at
depth >= 2. Same for `execute_n_hop` at 3+ rels.

#424 made this an error. This PR makes it work.

## Approach

`execute_n_hop` is already a general linear-chain traversal: it walks hop by
hop, carries a `HashMap<var, Value>` of accumulated bindings along each
in-progress path, evaluates WHERE over the full binding map at the end, and
projects `var.prop` columns. The only thing it could not do was take more than
one step per hop.

So each hop now checks for a quantifier. A varlen hop expands the frontier with
`execute_variable_hops` — the exact DFS/BFS the standalone
`execute_variable_length` path uses, including its simple-path constraint, its
`*0..n` handling and its `SAFETY_CAP` of 10 — instead of taking a single step.
Fixed hops are untouched. Projection, WHERE, DISTINCT, ORDER BY, SKIP and LIMIT
all come for free because they already operate on the accumulated bindings, not
on a hop-specific shape.

`execute_match` gains an `is_var_len_chain` test evaluated **before**
`is_two_hop`/`is_n_hop`, so relationship count can no longer silently outvote a
quantifier. The guard and its error message are removed.

Reachability-vs-enumeration follows the standalone path's rule (#165): a varlen
hop collapses to reachability BFS only when the query is `DISTINCT` and no
relationship variable is bound; otherwise it enumerates simple paths.

## Second bug found on the way — partial fix for #429

Making IC3 correct required fixing a separate silent-wrongness in the same
function, now filed as **#429**. `execute_n_hop` resolved relationship IDs by
**type name only** (`resolve_rel_ids_for_type`), while CSRs are keyed by
rel-table id and indexed by *source slot*. LDBC registers `isLocatedIn` twice —
`(Person)->(Place)` and `(Organisation)->(Place)` — so unioning both tables made
`Person` slot 2 inherit `Organisation` slot 2's edges. Concretely: IC3 for
Germany returned **Carol White**, who lives in the United States, because
TechStart Inc happened to occupy the same slot number.

Fixed hops now narrow to the single rel table for the `(src_label, dst_label,
rel_type)` triple — the discrimination `resolve_rel_table_id` already gives the
one- and two-hop executors. This also fixes plain 3+-hop queries that never
involved a quantifier.

**#429 stays open**: this only covers hops where *both* endpoint labels are
declared. Hops with an unlabeled intermediate node, and variable-length hops
(whose intermediate nodes are not constrained by the endpoint labels, so they
must keep the type-wide set), are still exposed. #429 also raises the broader
question of whether `csr_neighbors_filtered` should take a source label at all
rather than relying on every caller to pre-narrow. It is the same
slot-identity-treated-as-node-identity confusion as #427 and #415.

## What now works

- `(a)-[:R*m..n]->(b)-[:S]->(c)` — varlen plus one trailing fixed hop
- `(a)-[:R*m..n]->(b)-[:S]->(c)-[:T]->(d)` — varlen plus several trailing hops
- `(a)-[:S]->(b)-[:R*m..n]->(c)` — quantifier at any position, not just first
- more than one quantifier in the same chain
- `*0..n` (binds the source node itself), `*m..m`, `*m..`, `*`
- WHERE and inline property filters on any node in the chain, including the ones
  after the varlen hop
- DISTINCT, ORDER BY, SKIP, LIMIT across the join
- OPTIONAL MATCH of the shape — previously the guard's `InvalidArgument` was
  absorbed into a NULL row, turning a loud error back into silent wrongness
- the chunked pipeline does not intercept it: every `ChunkedPlan` variant is
  shape-gated on `min_hops.is_none()` or `rels.is_empty()`, and there is a test
  asserting the chunked engine returns the same rows

## Shapes still unsupported

These **error** rather than returning plausible-but-wrong rows:

- **Incoming varlen in a chain** — `(a)<-[:R*1..2]-(b)-[:S]->(c)`.
  `execute_variable_hops` walks outgoing edges only. Returns `Unimplemented`.
  The standalone `execute_variable_length` already rejected this the same way.
- **Varlen inside a pipeline MATCH stage** —
  `MATCH … WITH a MATCH (a)-[:R*1..2]->(b) …`. `execute_pipeline_match_hop`
  only ever steps once along `rels[0]` and ignores the quantifier — exactly
  #421's failure mode in a different executor. Now returns `InvalidArgument`
  naming #421. Implementing it there is a separate job; a guard is the honest
  interim.
- **Unlabeled chain source** — `execute_n_hop` requires a resolvable label on
  `nodes[0]` and returns `Unimplemented` without one. Pre-existing.
- **#431 — cross-label varlen on a checkpointed graph.** With an empty delta
  log, `get_node_neighbors_labeled` cannot recover a destination's label from
  CSR alone and falls back to the *source* label, so a varlen hop that crosses
  labels and is then filtered by a different destination label yields nothing.
  Pre-existing and identical in `execute_variable_length`, so this PR neither
  fixes nor worsens it; changing label recovery underneath both callers is a
  separate, separately-testable change. The #427 agent hit the same wall in
  `bfs_shortest_path` and worked around it by deriving the label from the rel
  table's `(src_label_id, dst_label_id)` — that is the right general fix and is
  written up in #431. Note this does **not** affect IC3/IC11: their varlen
  segment is Person→Person, and the label-crossing hops are fixed ones.

Two adjacent silent-truncation bugs were found but deliberately **not** fixed
here, since neither is varlen-specific:

- **#430** — `execute_pipeline_match_stage` drops hops `2..n` of any multi-rel
  pattern after a `WITH`; it only ever binds `nodes[0]` and `nodes[1]`, so the
  third node projects as `NULL` and the result reads like a complete answer to
  a shorter query.
- A multi-pattern `MATCH` whose patterns contain relationships
  (`MATCH (a)-[:R]->(b), (c)-[:S]->(d)`) falls through to `execute_scan`, which
  ignores relationships entirely. Not yet filed; noted here so it is not lost.

## Tests

`crates/sparrowdb-execution/tests/regression_421_varlen_plus_hop.rs` — 15 new
engine-level tests over a hand-built 6-person / 2-city fixture whose expected
values are spelled out in the module docs and derived by hand from the edge
lists. **Verified to fail against the pre-fix engine**: checking out
`engine/hop.rs` and `engine/scan.rs` from `25f3559` and re-running produces 10
failures, each showing exactly the depth-1 truncation #421 describes
(e.g. `left: [(1, 10), (4, 20)]` vs `right: [(1, 10), (2, 20), (4, 20), (5, 10)]`).

`crates/sparrowdb/tests/spa_111_ldbc_ic3_ic14.rs` — `ic3_friends_in_country` and
`ic11_job_referral` un-ignored, and both promoted from `!is_empty()` to exact
expected values. Two new tests pin the depth-2 cases #421 was dropping. Every
expectation is derived from the fixture CSVs, quoted in comments above the tests:

| query | expected | why |
|---|---|---|
| IC3(1, "Germany") | Dave Brown, Frank Miller | Germany is place 3; residents are 4, 6, 9. Alice's 1..2-hop set is {2,3,4,5,6,7}. Dave (4) is at depth 2, Frank (6) at depth 1. |
| IC3(1, "United Kingdom") | Eve Davis, Bob Jones | UK is place 2; residents 2, 5, 8. Bob (2) depth 1, Eve (5) depth 2. ORDER BY lastName. |
| IC11(1, "Germany") | Dave Brown | Germany ← org 3 (TechStart), whose only employee is person 4, at depth 2. |
| IC11(1, "United States") | Eve Davis, Bob Jones | US ← org 1 (Acme), employing 1, 2 and 5; 2 and 5 are in the friend set, Alice is not. |

The two `#[ignore]`d #421 tests in `crates/sparrowdb-bench/tests/ldbc_ic_test.rs`
(`ic3_friends_in_countries_returns_both_germans`, `ic11_job_referral`) also pass
now — confirmed with `--ignored` — but that file belongs to another change, so
the attributes are left for it to remove.

## Verification

All runs used an isolated `CARGO_TARGET_DIR=/Volumes/Dev/target-421`. The shared
target dir is being written by several agents concurrently and has already
produced one bogus green in this session — a test binary built from a different
worktree's source.

| command | result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --workspace --exclude sparrowdb-ruby --all-targets --keep-going -- -D warnings` | clean |
| `cargo test -p sparrowdb-execution --no-fail-fast` | 108 passed, 0 failed (11 binaries, incl. all 15 new #421 tests) |
| `cargo test -p sparrowdb --no-fail-fast` | 160 of 161 test binaries green, 0 failures (161st is #434, see below) |

**One target did not finish locally**: `spa_222_csr_lazy_load`. Its single test
builds 50 000 nodes and 100 000 edges through `execute_batch` in a debug build;
the local runner reaped it twice at ~49 and ~51 minutes, both SIGTERM rather than
failure. This is now accounted for independently rather than assumed: CI logs
show `spa222_mmap_open_and_traverse` takes **80 minutes and passes** on a green
`ubuntu-latest` run (03:37 start → 04:57 ok), i.e. one test is essentially the
whole ~85-minute CI wall-clock while the rest of the integration suite finishes
in minutes. Filed as #434. It is also unreachable from this change: its queries
are a two-node `MATCH (a:Person {id: i}), (b:Person {id: j}) CREATE (a)-[:KNOWS]->(b)`
— no relationships in the MATCH, so `execute_n_hop` is never entered — plus one
single-hop traversal.

Every target that could plausibly regress from this change **was** run and is
green, including `spa_252_three_hop_binding`, `spa_354_varlength_terminal_label`,
`spa_variable_paths`, `spa_224_varpath_reserved_label`, `spa_241_multihop_props`,
`spa_289_multi_label`, `spa_131_optional_match`, `spa_136_shortest_path`, and
`spa_299_{chunked_pipeline,phase2,phase3,phase4}_parity`.

### DIFF
--- a/crates/sparrowdb-execution/src/engine/hop.rs
+++ b/crates/sparrowdb-execution/src/engine/hop.rs
@@ -1737,6 +1737,19 @@ impl Engine {
     ///
     /// This replaces the previous fallthrough to `execute_scan` which only
     /// scanned the first node and ignored all relationship hops.
+    ///
+    /// # Variable-length hops (#421)
+    ///
+    /// Any hop in the chain may carry a variable-length quantifier
+    /// (`-[:R*1..2]->`).  Such a hop expands the frontier with
+    /// [`Engine::execute_variable_hops`] — the same DFS/BFS used by the
+    /// standalone `execute_variable_length` path — instead of taking a single
+    /// step, so `(a)-[:R*1..2]->(b)-[:S]->(c)` binds `b` to every node reachable
+    /// from `a` in 1..2 hops and then joins each of those against the trailing
+    /// `:S` hop.  Before this, `is_var_len` in `execute_match` only recognised a
+    /// quantifier when the varlen relationship was the *only* relationship in
+    /// the pattern, so a trailing hop silently demoted `*1..2` to a plain single
+    /// hop and every depth >= 2 match was dropped without an error.
     pub(crate) fn execute_n_hop(
         &self,
         m: &MatchStatement,
@@ -1810,10 +1823,61 @@ impl Engine {
         // We read all delta edges once up front to avoid repeated file I/O.
         let delta_all = self.read_delta_all();
 
+        // ── #421: variable-length quantifiers anywhere in the chain ──────────
+        //
+        // `Some((min, max))` for a hop written `-[:R*min..max]->`.  An open
+        // upper bound is capped at 10, matching `execute_variable_length`.
+        let varlen_per_hop: Vec<Option<(u32, u32)>> = pat
+            .rels
+            .iter()
+            .map(|r| {
+                if r.min_hops.is_some() || r.max_hops.is_some() {
+                    Some((r.min_hops.unwrap_or(1), r.max_hops.unwrap_or(10)))
+                } else {
+                    None
+                }
+            })
+            .collect();
+        let has_varlen = varlen_per_hop.iter().any(Option::is_some);
+
         // Pre-resolve per-hop rel-table IDs so the inner loop uses filtered
         // CSR lookups instead of scanning every relation type (SPA-284).
+        //
+        // #429: a relationship *type* can back several rel tables — LDBC's
+        // `isLocatedIn` exists as both `(Person)->(Place)` and
+        // `(Organisation)->(Place)`.  CSRs are keyed by rel-table id and indexed
+        // by *source slot*, so unioning every table of a given type makes
+        // `Person` slot 2 inherit `Organisation` slot 2's edges.  (Concretely:
+        // Carol White, Person slot 2, appeared to live in Germany because
+        // TechStart Inc is Organisation slot 2.)  When both endpoint labels of a
+        // fixed hop are declared, narrow to the single rel table for that triple
+        // — the same discrimination `resolve_rel_table_id` already gives the
+        // one- and two-hop executors.
+        //
+        // Two cases stay exposed and are tracked in #429: hops whose endpoint
+        // labels are not both declared, and variable-length hops, whose
+        // intermediate nodes are not constrained by the endpoint labels and so
+        // must keep the type-wide set.
         let rel_ids_per_hop: Vec<Vec<u32>> = (0..n_rels)
-            .map(|i| self.resolve_rel_ids_for_type(&pat.rels[i].rel_type))
+            .map(|i| {
+                let type_wide = self.resolve_rel_ids_for_type(&pat.rels[i].rel_type);
+                if varlen_per_hop[i].is_some() || type_wide.len() < 2 {
+                    return type_wide;
+                }
+                match (label_ids_per_node[i], label_ids_per_node[i + 1]) {
+                    (Some(src), Some(dst)) => self
+                        .snapshot
+                        .catalog
+                        .get_rel_table(src as u16, dst as u16, &pat.rels[i].rel_type)
+                        .ok()
+                        .flatten()
+                        .map(|id| vec![id as u32])
+                        // No table for this (src, dst, type) triple: the hop
+                        // cannot match anything.
+                        .unwrap_or_default(),
+                    _ => type_wide,
+                }
+            })
             .collect();
         // If any hop specifies a rel type that doesn't exist in the catalog, no
         // traversal can produce results — return empty immediately.
@@ -1825,6 +1889,33 @@ impl Engine {
                 });
             }
         }
+
+        // `execute_variable_hops` only walks outgoing edges.  Reject rather than
+        // quietly answering an `<-[:R*1..2]-` pattern with the wrong direction.
+        for (i, v) in varlen_per_hop.iter().enumerate() {
+            if v.is_some() && pat.rels[i].dir != sparrowdb_cypher::ast::EdgeDir::Outgoing {
+                return Err(sparrowdb_common::Error::Unimplemented);
+            }
+        }
+
+        // Delta index + label registry used by the variable-length expansion.
+        // Built once (not per source node) and only when a varlen hop exists.
+        let (delta_idx, node_label, all_label_ids) = if has_varlen {
+            let idx = build_delta_index(&delta_all);
+            let mut labels: HashSet<(u64, u32)> = HashSet::new();
+            for r in &delta_all {
+                labels.insert((r.src.0 & 0xFFFF_FFFF, (r.src.0 >> 32) as u32));
+                labels.insert((r.dst.0 & 0xFFFF_FFFF, (r.dst.0 >> 32) as u32));
+            }
+            let mut ids: Vec<u32> = labels.iter().map(|&(_, l)| l).collect();
+            ids.sort_unstable();
+            ids.dedup();
+            (idx, labels, ids)
+        } else {
+            (DeltaIndex::new(), HashSet::new(), Vec::new())
+        };
+        // Reusable neighbor buffer for the variable-length traversal.
+        let mut neighbors_buf: HashSet<(u64, u32)> = HashSet::new();
 
         let mut rows: Vec<Vec<Value>> = Vec::new();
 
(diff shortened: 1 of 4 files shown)


ryaker:
@coderabbitai review

ryaker:
@coderabbitai review

Previous attempts on this PR returned "Review skipped: reviews are disabled for this base branch" (it was stacked on a sibling branch) and later "Review limit reached" — both with a green check and no review. The PR now targets `main` and has been rebased, so requesting a real review of the current diff.

ryaker:
Closing as fully subsumed by #432.

Verification: `git diff c9f7a0b 8791663 --stat` (this branch's own merge-base with main, to its tip) shows exactly 4 files (hop.rs, scan.rs, and its two test files), 701 insertions / 101 deletions. The equivalent commit inside #432 — `feat(engine): variable-length path followed by further hops (closes #421)` — has the identical stat: same 4 files, same 701/101 line counts. Its two test files (`regression_421_varlen_plus_hop.rs`, `spa_111_ldbc_ic3_ic14.rs`) are byte-identical between this branch's worktree and #432's.

#432 also fixed a real bug in this code (frontier dropping the resolved label on each hop advance, `unwrap_or(0)` treating a real label id as a sentinel) that CodeRabbit found during review — so #432 is not just "contains this PR's diff," it's a corrected superset.

Not merging this branch directly since #432 already carries an improved version of the same commit.

Issue #434, opened 2026-08-02
spa222_mmap_open_and_traverse takes 80 minutes and dominates every CI run


ryaker:
## Summary

A single test, `spa222_mmap_open_and_traverse` in `crates/sparrowdb/tests/spa_222_csr_lazy_load.rs`, runs for **80 minutes**. It accounts for essentially the whole CI wall-clock — the rest of the integration suite finishes in a few minutes.

Measured on a green `Test (ubuntu-latest)` run:

```
03:37:00  Running tests/spa_222_csr_lazy_load.rs
03:38:00  test spa222_mmap_open_and_traverse has been running for over 60 seconds
04:57:30  test spa222_mmap_open_and_traverse ... ok
```

It passes. The problem is cost, not correctness.

## Why it matters now

Every PR waits ~85 minutes for a verdict that could arrive in ~5. During the 2026-08-01 bug sweep this serialised the whole merge chain, and it repeatedly caused agents to report a test as "hung" when it was merely starved of CPU by concurrent builds — one agent made three attempts and never saw it finish locally.

## Cause

The test issues roughly 150,000 sequential queries in a debug build: ~50,000 `CREATE` statements, then ~100,000 `MATCH` + `CREATE` where both endpoints are resolved by property lookup. Per-query overhead dominates; the thing under test (CSR lazy load / mmap open) is not what the time is spent on.

## Options

1. **Shrink the fixture.** Does lazy CSR load need 50K nodes / 100K edges to be exercised, or would 5K/10K cover the same code path? Likely a 10x cut for no loss of coverage.
2. **Seed in bulk.** Build the graph through the loader or `WriteTx` rather than one Cypher statement at a time — the traversal is the subject, not the ingest path.
3. **Mark `#[ignore]` and run it on a schedule**, or gate it behind a `--features slow-tests` flag, so PR CI stays fast and the coverage still runs nightly.

Option 1 or 2 is preferable; option 3 keeps coverage but removes it from the PR signal.

## Related

Surfaced while verifying #427, from that agent's inability to see the test complete locally. Adjacent to #412, which made CI actually gate — a gate that takes 85 minutes gets worked around.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Aq3C5J646sBCH7sVs3uaCE

Issue #419, opened 2026-08-01
Multi-label / secondary-label support does not extend to relationship patterns (both engines)


ryaker:
## Summary

SPA-289 Phase 1 implemented multi-label support for **node scans only**. Relationship patterns ignore it entirely, in **both** the chunked pipeline and the row-at-a-time engine.

Surfaced by CodeRabbit review on #416, then verified by writing the tests.

## Reproduction

```rust
CREATE (n:Animal:Pet {name: 'Rex'})
CREATE (n:Animal {name: 'Simba'})
CREATE (n:Person {name: 'Ann'})
MATCH (a:Person {name:'Ann'}),(b:Animal {name:'Rex'})  CREATE (a)-[:OWNS]->(b)
MATCH (a:Person {name:'Ann'}),(b:Animal {name:'Simba'}) CREATE (a)-[:OWNS]->(b)
```

| Query | Expected | Actual |
|---|---|---|
| `MATCH (a:Person)-[:OWNS]->(b:Pet) RETURN a.name` | 1 row | **0 rows** — secondary-label endpoint invisible |
| `MATCH (a:Person)-[:OWNS]->(b:Animal:Pet) RETURN b.name` | 1 row (Rex) | **2 rows** (Rex, Simba) — labels after the first ignored |

## Why #416 does not fix this

#416 adds `all_chunked_node_labels_are_primary_only`, which gates **every** chunked plan (Scan, OneHop, TwoHop, MutualNeighbors) so multi-label and secondary-label patterns fall back to the row engine. That is correct and necessary for the node-scan case.

But it cannot fix relationship patterns, because **the row engine is equally unimplemented here**. The SPA-200 secondary-label union and `execute_scan_multi_label` intersection both live in the single-node scan path only. Falling back just routes to a different wrong answer.

## Work required

- Row engine: teach the relationship/traversal path to resolve endpoints via `resolve_multi_label_node_ids` rather than a primary-label scan.
- Chunked plans: either the same, or keep them gated (as #416 does) and accept the fallback.

## Tests

Two `#[ignore]`d regression tests land with #416 in `crates/sparrowdb/tests/spa_289_multi_label.rs`, pointing at this issue:

- `one_hop_match_honours_secondary_label_endpoint`
- `one_hop_match_honours_multi_label_endpoint`

Remove the `#[ignore]` when implementing.

## Related

- #415 / #416 — node-scan multi-label fix
- #330 — original SPA-289 Phase 1

[...]

PR #437, opened 2026-08-02
fix(engine): neighbour lookups carry (slot, label) and read labels from the catalog (#429, #431)

ryaker:
## The one bug behind four tickets

A `NodeId` is `(label_id << 32) | slot`, so slot 2 exists independently in
`Person`, `Place`, `Org` and every other label. `csr_neighbors_filtered` and
`csr_neighbors_all` took a **bare slot with no label** and returned **bare
slots**:

```rust
fn csr_neighbors_filtered(&self, src_slot: u64, rel_ids: &[u32]) -> Vec<u64>
```

The signature made it impossible for a caller to be correct, and every caller
inherited both halves of the defect:

1. **Reading another label's edges.** A `CsrForward` is indexed by the source
   slot of *one* relationship table. Probing every table with a bare slot lets
   `Person` slot 2 inherit `Org` slot 2's edges (#429).
2. **Losing the destination's label.** Callers had to reconstruct it.
   `get_node_neighbors_labeled` did so from delta-log hints, falling back to the
   **source's** label. `checkpoint()` truncates the delta log, so after a
   checkpoint every hint vanished and every CSR neighbour silently took the
   source's label (#431).

## The fix

The catalog already registers every relationship table as
`(id, src_label_id, dst_label_id, rel_type)`, and a CSR holds exactly one
table's edges. So for a CSR hit the destination label is **known**, and a table
whose `src_label_id` differs is not indexed by this node's slot at all.

```rust
fn csr_neighbors_labeled(&self, src_slot: u64, src_label_id: u32, rel_ids: &[u32])
    -> Vec<(u64, u32)>   // (dst_slot, dst_label_id)
```

plus `csr_neighbor_slots_to_label(..)` for callers that already know the
destination label. Both old helpers are gone. A lazily-built `CsrTopology` on
`ReadSnapshot` caches the label triples per snapshot —
`list_rel_tables_with_ids()` clones a `String` per table and is called inside
traversal loops.

Call sites:

| site | change |
|---|---|
| `get_node_neighbors_labeled` (path.rs) | drops the `node_label` / `all_label_ids` delta-hint parameters entirely |
| `execute_n_hop` (hop.rs) | frontier carries `(slot, label_id)` instead of re-deriving the label from the pattern; each hop filters destinations on the label; unlabeled intermediates are addressed by their real label, not label 0 |
| `bfs_shortest_path` (expr.rs) | drops the catalog loop it grew in #432 and calls the shared helper |
| `RelTableLookup::All` arms in expr.rs / scan.rs / mutation.rs | restrict to tables running from this node's label to the destination label they build the `NodeId` from; the paired delta filters gain the same destination-label check |

## What this closes

* **#431 — closed.** Neighbour labels come from the catalog; the delta-hint
  fallback is deleted, not guarded. Nothing about the answer now depends on
  whether the graph has been checkpointed.
* **#429 — closed**, including both parts the ticket left open: undeclared
  endpoint labels (the label is carried, so there is nothing to declare) and
  `execute_variable_hops` / `execute_variable_length` (they go through the same
  helper).
* **#427 — already closed by #432**; this replaces that fix's local catalog loop
  with the shared helper, which is now the only implementation. The #427
  regression suite still passes unchanged.
* **#415 — not addressed.** It is the same *family* of confusion but a different
  mechanism: the chunked pipeline resolves nodes by `labels[0]` against that
  label's own primary store. That is a node-scan path, not a neighbour lookup;
  the gating from #416 remains necessary and #419 tracks the row engine's
  matching gap.

## Test

`crates/sparrowdb/tests/regression_slot_identity_root_cause.rs` — four node
labels, three relationship types across **four** relationship tables
(`LOCATED_IN` spans `(Person)->(Place)` and `(Org)->(Place)`), deliberate
cross-label slot collisions (Person 2 = Carol → Kyoto, Org 2 = TechStart →
Berlin), asserted on **both sides of `checkpoint()`**. Every expected value is
derived by hand in the test comments; none was captured from program output.

**Measured against the pre-fix engine files** (engine sources checked out at the
base commit, test file unchanged, then restored byte-exact — sha256 verified):

```
test result: FAILED. 3 passed; 4 failed
  three_hop_chain_does_not_inherit_another_labels_edges     ["Berlin","Kyoto"] vs ["Kyoto"]   (#429)
  unlabeled_intermediates_do_not_inherit_another_labels_edges ["Berlin","Kyoto"] vs ["Kyoto"] (#429)
  cross_label_varlen_survives_checkpoint                    [] vs ["Springfield"]             (#431)
  cross_label_varlen_does_not_read_a_colliding_slot         [] vs ["Kyoto"]                   (#431)
```

The 3 that pass pre-fix are deliberate controls: homogeneous varlen, one-hop,
and shortestPath (already fixed by #432).

## Verification

* `cargo test -p sparrowdb-execution --no-fail-fast` — 10 binaries, all ok
* `cargo test -p sparrowdb --no-fail-fast` — 163 binaries, **1033 passed, 0
  failed, 10 ignored**, excluding `spa_222_csr_lazy_load` (~80 min runtime,
  being addressed separately in #434)
* `cargo fmt --all -- --check` — clean, re-verified in a fresh clone of the
  pushed branch
* `cargo clippy --workspace --exclude sparrowdb-ruby --all-targets --keep-going
  -- -D warnings` — clean

No existing test changed behaviour; nothing was weakened.

## Branching note

Requested base is `fix/issue-427-...` (#432) "on top of both #432 and #433". In
git, #432's branch is **not** a descendant of #433's — both fork from `25f3559`
— so this sits on #432 only. It does not contain #433's `execute_n_hop` changes,
and since this rewrites that function's frontier, expect a conflict there when
#433 lands. The root fix subsumes what #433 does for `execute_n_hop` (narrowing
the rel-table set when both endpoint labels are declared), so the resolution
should be to take this version.

Refs #415 #427 #429 #431

### DIFF
--- a/crates/sparrowdb-execution/src/engine/expr.rs
+++ b/crates/sparrowdb-execution/src/engine/expr.rs
@@ -408,7 +408,13 @@ impl Engine {
         let csr_nb: Vec<u64> = match rel_lookup {
             RelTableLookup::Found(rtid) => self.csr_neighbors(rtid, src_slot),
             RelTableLookup::NotFound => return false,
-            RelTableLookup::All => self.csr_neighbors_all(src_slot),
+            // Untyped edge: every table whose source label is this node's, and
+            // whose destination label is the one the pattern asks for.  A bare
+            // slot would otherwise let a neighbour of any label answer for one
+            // of `dst_label_id_opt`.
+            RelTableLookup::All => {
+                self.csr_neighbor_slots_to_label(src_slot, src_label_id, dst_label_id_opt, &[])
+            }
         };
         let delta_nb: Vec<u64> = self
             .read_delta_all()
@@ -603,28 +609,8 @@ impl Engine {
         // SPA-283: build HashMap index for O(1) per-node delta lookups.
         let delta_idx = build_delta_index(&delta_all);
 
-        // #427: derive CSR neighbour labels from the catalog rather than guessing.
-        //
-        // A relationship table is registered as `(rel_table_id, src_label_id,
-        // dst_label_id, rel_type)` and its CSR holds only that table's edges.  So
-        // for a CSR hit we know the destination's label exactly — it is the
-        // table's `dst_label_id` — and we know the table only applies to sources
-        // of its `src_label_id`.  Both matter: the shared
-        // `get_node_neighbors_labeled` helper instead falls back to the *source's*
-        // label whenever the delta log carries no hint, which is always true once
-        // a graph has been checkpointed, and it probes every CSR with the raw slot
-        // regardless of the source's label.
-        let rel_tables: Vec<(u32, u32, u32)> = self
-            .snapshot
-            .catalog
-            .list_rel_tables_with_ids()
-            .into_iter()
-            .filter(|(id, _, _, _)| rel_ids.is_empty() || rel_ids.contains(&(*id as u32)))
-            .map(|(id, s_lid, d_lid, _)| (id as u32, s_lid as u32, d_lid as u32))
-            .collect();
-
         // Frontier carries (slot, label_id) so each hop uses the correct label
-        // when probing the delta index.
+        // when looking up neighbours.
         let mut visited: std::collections::HashSet<(u64, u32)> = std::collections::HashSet::new();
         visited.insert((src_slot, src_label_id));
         let mut frontier: Vec<(u64, u32)> = vec![(src_slot, src_label_id)];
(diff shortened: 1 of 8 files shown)


ryaker:
Closing as fully subsumed by #432.

Verification, file by file, comparing this branch's worktree against #432's:
- `mod.rs`, `path.rs`, `mutation.rs`: byte-identical (`diff` exit 0).
- `expr.rs`: differs only by unrelated `hybrid_search` caching work (#456) that landed on `main` after this branch's original commit was written — the actual markers this fix introduced (`csr_neighbor_slots_to_label`, `get_node_neighbors_labeled` in `bfs_shortest_path`, `csr_neighbors_all` removed) are all present in #432.
- `hop.rs`: byte-identical once this worktree's own uncommitted follow-up fix (dropping the now-unused `node_label`/`all_label_ids` params from `execute_variable_hops` after #431's signature simplification) is included — that follow-up was never committed here before I moved to #432, where it's committed as part of `fix(engine): neighbour lookups carry (slot, label) and read labels from the catalog`.
- `regression_slot_identity_root_cause.rs`: #432's copy is a strict superset — same 6 original tests (trivial comment rewording only), plus 2 more (`varlen_plus_trailing_hop_does_not_read_a_colliding_slot` and `varlen_plus_trailing_hop_unlabeled_destination_does_not_default_to_label_zero`) added in response to CodeRabbit review on #432, both confirmed to fail against #421's commit alone.

The label-carrying fix in `execute_n_hop` (`hop.rs`) that this branch's original commit did *not* yet cover (it was written before #421's varlen dispatch existed) was completed directly on #432 once CodeRabbit found the gap — so #432 is a corrected superset, not merely a rebase of this content.

Issue #396, opened 2026-04-04
feat(hybrid-search): RRF fusion of vector + BM25 results in single Cypher query


ryaker:
## Why

Vector search (#394) finds semantically similar content. Full-text search (#395) finds exact keyword matches. Neither alone is sufficient for production RAG — the best recall comes from fusing both. This is the \"hybrid search\" pattern used by Pinecone, Weaviate, and every serious vector DB.

SparrowDB can do this in a single Cypher query with no external services, which is the entire point.

## Depends On

- #394 (HNSW vector index)
- #395 (BM25 full-text index)

### Query function
\`\`\`cypher
MATCH (n:Memory)
WHERE hybrid_search(n, 'content', \$query_vec, 'transformer attention', { alpha: 0.7 })
RETURN n ORDER BY hybrid_score(n) DESC LIMIT 10
\`\`\`

Or via explicit fusion:
\`\`\`cypher
WITH vector_search('Memory', 'embedding', \$vec, 50) AS vec_results,
     full_text_search('Memory', 'content', 'transformer', 50) AS fts_results
WITH rrf_fusion(vec_results, fts_results) AS fused
UNWIND fused AS row
MATCH (n) WHERE id(n) = row.node_id
RETURN n, row.score ORDER BY row.score DESC LIMIT 10
\`\`\`

### Fusion functions
| Function | Description |
|----------|-------------|
| \`rrf_fusion(list1, list2, k=60)\` | Reciprocal Rank Fusion: \`1/(k+rank1) + 1/(k+rank2)\` |
| \`weighted_fusion(list1, list2, alpha)\` | \`alpha * norm_vec_score + (1-alpha) * norm_bm25_score\` |
| \`hybrid_score(n)\` | Returns fused score set by most recent hybrid_search call |

### Planner optimization
- Detect \`hybrid_search()\` → single HybridScan plan node that runs both indexes in parallel (rayon), then fuses
- Avoids materializing full lists when LIMIT is present (early termination)

## Acceptance Criteria

- [ ] \`rrf_fusion()\` produces correct merged ranking from two scored lists
- [ ] \`weighted_fusion()\` with alpha parameter works
- [ ] \`hybrid_search()\` convenience function runs both indexes and fuses
- [ ] LIMIT pushdown: stops scanning after enough candidates found
- [ ] Result quality test: hybrid beats pure vector and pure BM25 on recall@10 with mixed queries
- [ ] MCP tool \`hybrid_search\` exposed

## Notes

[...]

Issue #395, opened 2026-04-04
feat(fts): BM25 full-text index + CONTAINS/STARTS WITH acceleration


ryaker:
## Why

Knowledge retrieval needs text search. Right now \`WHERE n.content CONTAINS 'transformer'\` does a full scan. For a KMS with thousands of memory nodes, that's too slow. A BM25 inverted index enables ranked full-text retrieval and accelerates the substring predicates already in Cypher.

Combined with vector search (#394), this enables hybrid retrieval — the foundation of modern RAG pipelines — without leaving SparrowDB.

### Index management (DDL)
\`\`\`cypher
CREATE FULLTEXT INDEX FOR (n:Memory) ON (n.content)
CREATE FULLTEXT INDEX FOR (n:Document) ON (n.title, n.body)
DROP INDEX memory_content_fts
\`\`\`

### Query functions (Cypher)
| Function | Returns | Notes |
|----------|---------|-------|
| \`full_text_search(label, prop, 'query string')\` | List<{node, score}> | BM25 ranked results |
| \`bm25_score(prop, 'query')\` | Float64 | score usable in WHERE/ORDER |

Example:
\`\`\`cypher
MATCH (n:Memory)
WHERE full_text_search('Memory', 'content', 'transformer attention mechanism')
RETURN n, bm25_score(n.content, 'transformer attention mechanism') AS score
ORDER BY score DESC LIMIT 20
\`\`\`

### Existing predicate acceleration
- \`WHERE n.prop CONTAINS 'x'\` → bitmap filter from posting list before full scan
- \`WHERE n.prop STARTS WITH 'x'\` → prefix lookup in trie/sorted terms

### Storage
- Inverted index: \`HashMap<term, PostingList>\` where \`PostingList = Vec<(NodeId, tf)>\`
- Persisted to \`<db>/fts_indexes/<label>_<prop>.fts\` (bincode)
- BM25 parameters: \`k1 = 1.2\`, \`b = 0.75\` (tunable via index OPTIONS)
- Term normalization: lowercase, unicode-aware tokenization, optional stemming

### Tokenizer
- Default: whitespace + punctuation split, lowercase
- Optional: porter stemmer (pure Rust, no deps)
- Store term → IDF at index build time; TF per document in posting list

### Planner
- Detect \`full_text_search()\` calls → FTSScan plan node
- Detect \`CONTAINS\`/\`STARTS WITH\` on indexed properties → bitmap pre-filter

## Implementation Notes

[...]

PR #438, opened 2026-08-02
docs(ws5): FUTURE-vector-bridge — docs-only deliverable of a deferred workstream

ryaker:
## This PR contains no code, and that is the deliverable

WS5 of the Sparrow integrations directive is **explicitly deferred**: "Do not implement in
this cycle. Produce only `docs/FUTURE-vector-bridge.md` … One page. Stop there." So there is
nothing to review for correctness beyond the doc's factual claims. No reviewer should expect
a vector bridge here.

**One file added:** `docs/FUTURE-vector-bridge.md`. Nothing else is touched.

## Why deferred (the directive's reasoning, recorded in the doc)

WS1–WS4 each make Sparrow fit into someone else's stack. A vector feature makes Sparrow
*bigger*, which is the opposite bet, and it competes on an axis with brutal incumbents.
Revisit after WS2 and WS3 have users.

## What the doc actually says

It is not a menu of options. Every claim was checked against `main` @ `74c1219` and cites a
file path or issue number:

- **The premise is mostly shipped already.** HNSW (#394/#398), BM25 (#395/#399), and
  RRF/weighted fusion (#396/#403) are in the tree, plus Node-binding `vectorSearch` /
  `hybridSearch`. So the doc covers what is *missing* rather than proposing to "add vectors."
- **The DDL entry point never worked.** `CREATE VECTOR INDEX` was unparseable until PR #420
  (#417, commit `25f3559`, merged today). Every published release through v0.1.21 shipped the
  feature unreachable from Cypher, and `docs/` still has zero coverage of it. That matters
  for how much demand signal to read into the feature.
- **The gaps are lifecycle, not algorithm:** re-embedding an existing node is a silent no-op
  (`vector_index.rs:374`); there is no delete; `save()` is a non-atomic whole-index
  `fs::write`; `hybrid_search` reloads from disk per call and ignores the in-memory index;
  and no test joins `hybrid_search` output back to nodes.
- **API sketch** for `similar(label, prop, vector, k)` with the joinable Cypher form, plus an
  explicit warning that returned ids must be full `NodeId`s — `(label_id << 32) | slot`.
  Treating a slot as identity has produced four separate silent-wrong-rows bugs here (#415,
  #427, #429, #431), and the `NodeId` doc comment in `sparrowdb-common` is itself stale
  (`<< 48`), so an implementer trusting it would pack ids wrong.
- **Judgement on the open question** (do embeddings belong in the storage layer): sidecar,
  rebuildable, not WAL-backed — because the vector is already a node property, making the
  HNSW graph derived state in the same category as the property and text indexes, which
  `property_index.rs:561` already treats as rebuildable from column files.
- **Judgement on in-process vs. delegated vector store:** keep in-process HNSW; do not build
  a pluggable-backend abstraction. Delegation cannot fuse with BM25 over the same `NodeId`
  space without crossing a serialization boundary, and it would introduce a required network
  dependency that ground rule §2.6 forbids. Users with an existing pgvector/Pinecone are
  served by a docs page in WS3's LangChain layer, which needs no SparrowDB code.

## Verification

Markdown only — nothing to compile, no CI-relevant change. Formatted with `prettier`.

## Summary by CodeRabbit

* **Documentation**
  * Added documentation outlining future plans for vector-bridge capabilities.
  * Cataloged existing search, persistence, and binding functionality.
  * Documented potential APIs, identity requirements, storage approaches, and an ordered backlog for future consideration.

### DIFF
--- a/docs/FUTURE-vector-bridge.md
+++ b/docs/FUTURE-vector-bridge.md
@@ -225,11 +225,21 @@ Note that "deferred" here does not mean "no vector code" — the substrate alrea
 means no new vector _surface area_. The highest-value use of the deferral is to make what
 exists correct rather than to extend it. In order:
 
-1. Close out #400 and document HNSW / `hybrid_search` in `docs/` — today there is zero
+1. **Correct the authoritative `NodeId` encoding first (#474)** — this gates everything below
+   that crosses the index boundary. Three contracts currently disagree: production packs
+   `label_id << 32`, while the `NodeId` doc comment and `node_id_packing_roundtrip` both
+   describe `<< 48`, and `VectorIndex::insert` takes a bare `u64` that accepts either. Adding
+   a `VectorKey(NodeId)` newtype only at the `similar()` boundary would leave every existing
+   caller free to pass a raw slot. Fix the comment, rewrite the test to assert against the
+   production packing helper rather than re-implementing the shift, then tighten the
+   `VectorIndex` boundary and migrate its callers. A slot and a full `NodeId` are both `u64`,
+   so getting this wrong is a silent wrong-answer bug rather than a type error — the same
+   root cause as #427, #415, and the `execute_n_hop` cross-label edge bug.
+2. Close out #400 and document HNSW / `hybrid_search` in `docs/` — today there is zero
    user-facing coverage of a shipped feature.
-2. Upsert and delete semantics in `VectorIndex` (§3 items 3–4). Silent stale results are worse
+3. Upsert and delete semantics in `VectorIndex` (§3 items 3–4). Silent stale results are worse
    than a missing feature.
-3. Atomic `save()` (temp file + rename) and the LSN stamp from §5.
-4. Make `hybrid_search` use the in-memory index instead of reloading from disk (§3 item 6).
-5. Only then `similar()` in Cypher — landing with an integration test that actually joins its
-   output back to nodes, which no test does today.
+4. Atomic `save()` (temp file + rename) and the LSN stamp from §5.
+5. Make `hybrid_search` use the in-memory index instead of reloading from disk (§3 item 6).
+6. Only then `similar()` in Cypher — landing with an integration test that actually joins its
+   output back to nodes, which no test does today, and with §7's three questions answered.
--- a/docs/FUTURE-vector-bridge.md
+++ b/docs/FUTURE-vector-bridge.md
@@ -1,0 +1,191 @@
+# FUTURE: Hybrid vector bridge (WS5 — deferred)
+
+**Status:** deferred, deliberately. This page is the whole deliverable of WS5 in the Sparrow
+integrations directive. No vector code is to be written against it.
+
+**Why deferred, in the directive's own terms:** WS1–WS4 each make Sparrow fit into someone
+else's stack. A vector feature makes Sparrow _bigger_, which is the opposite bet, and it
+competes on an axis with brutal incumbents. Revisit after WS2 (`sparrow-vault`) and WS3
+(`langchain-sparrowdb`) have users.
+
+Written 2026-08-01 against `main` @ `74c1219`. Every claim below cites a file or an issue so
+a later reader can check whether it is still true.
+
+---
+
+## 1. The premise is mostly already shipped
+
+WS5 was written as "pair SparrowDB's typed graph with an embedding index." That pairing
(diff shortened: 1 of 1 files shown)


ryaker:
@coderabbitai review

The earlier attempt hit the account PR review limit, so the check went green without a review having happened. Requesting a real one.

ryaker:
@coderabbitai full review

Rebased onto `6254f91`. The previous CodeRabbit run returned `Review rate limited` — no review actually happened, so this is the first real pass. Docs-only: a single new file, `docs/FUTURE-vector-bridge.md`, describing a deferred workstream.

ryaker:
@coderabbitai full review

Retrying: the previous request returned `Action not completed — Review rate limited` inside a collapsed details block, so no review has run on this PR yet.

ryaker:
Addressed in `2900031`. Thanks — the review was substantive and one finding led somewhere well outside this PR.

**Quick wins, all fixed:**
- The self-invalidating grep. You were right that it becomes false the moment this file merges. Rewritten as a console block with `--exclude=FUTURE-vector-bridge.md` so the claim survives its own merge.
- MD040 on the API-signature fence — now `text`.
- Dangling section refs. `§3` has numbered items, not subsections, so `§3.3`/`§3.4` became `§3 items 3–4`. You caught two; there was a third, `§3.6` → `§3 item 6`. Also disambiguated `§2.6`, which reads as an internal ref but points at the integration directive.

**The three heavy-lift findings** — I did not invent answers for them, because this document's purpose is to record what a deferred workstream has not yet decided. New **§7** states all three as prerequisites that must be answered *before* `similar()` is specified, since each changes the API's shape:

1. **Score direction.** This one was a genuine error, not just an underspecification: `ORDER BY hit.score DESC` assumes higher-is-better, and Euclidean returns a distance where lower is more similar. Sorting it descending returns the *least* similar rows first, and `lib.rs:1123-1181` already supports that metric. Flagged inline at the example as well as in §7.
2. **Stale reads and index compatibility.** Adopted your framing directly — an LSN establishes staleness but not compatibility; dimension, metric, and format can all differ while the LSN looks current. §7 now requires both the during-rebuild query behaviour and the rejection metadata before the claim can be made.
3. **External ID mapping.** Recorded with the deleted/never-existed case you did not raise but which falls out of it — returning fewer than `k` rows silently is a different failure from raising.

**The NodeId finding was the most valuable thing here and I have filed it as #474.** You flagged three conflicting contracts. I checked: `grep -rn '<< 48' crates/` returns exactly one hit, the unit test `node_id_packing_roundtrip`, and the `NodeId` doc comment says *"upper 16 bits = label_id, lower 48 bits = slot_id"* — while all production code uses `<< 32`. The test passes because it packs and unpacks with its own shift, so it is self-consistent and would pass at any width. That matters beyond this PR: slot-vs-`NodeId` confusion is the confirmed root cause behind #427, the `execute_n_hop` cross-label bug, and #415. §7 now cross-references it.

ryaker:
Addressed in `06a513e`. You were right that this one wasn't closed — I had cross-referenced #474 from §7's open questions, but cross-referencing a prerequisite is not the same as scheduling it, and your finding was specifically about the prioritized order.

The `NodeId` encoding correction is now **item 1** in §8's list, ahead of everything that crosses the index boundary, with the reasoning you gave: a `VectorKey(NodeId)` newtype introduced only at the `similar()` boundary still leaves every existing `VectorIndex::insert` caller free to pass a raw slot, because a slot and a full `NodeId` are both `u64`. The scope stated there is the one you asked for — fix the comment, rewrite the test against the production packing helper rather than a re-implemented shift, tighten the boundary, migrate callers.

Also renumbered the rest of the list and made item 6 (`similar()` in Cypher) explicitly conditional on §7's three questions being answered, so the ordering constraint is visible from the backlog rather than only from the prose.

Worth saying plainly: the three-conflicting-contracts observation was the most valuable finding in this review, and it is about the codebase rather than this document. It is filed as #474 with the verification — `grep -rn '<< 48' crates/` returns exactly one hit, the unit test, which passes only because it packs and unpacks with its own shift and would pass identically at any width.

ryaker:
@coderabbitai review

`06a513e` addresses the remaining thread — the `NodeId` encoding correction is now item 1 in §8's prioritized backlog rather than only cross-referenced from §7. Please re-check that thread specifically; your last pass predates the push.

PR #442, opened 2026-08-02
fix(vector): stop silent embedding loss in the HNSW index (closes #441)

### LINKED ISSUE #441: Silent vector loss: HNSW index lost-update on concurrent opens, non-atomic save, silent re-insert no-op, unrecoverable embeddings
## Incident

A production consumer (KMSmcp) lost embeddings from `vector_indexes/hnsw_Knowledge_embedding.bin` with no error raised at any point.

| moment | file size | vectors |
|---|---|---|
| before an embedding backfill | 2,497,757 B | — |
| immediately after the backfill (reported 1721 written, 234 failed, 217 s) | 8,084,693 B | 2433 |
| after the hosting launchd daemon restarted | 4,352,529 B | **1299** |

The 4.35 MB file is a **complete, valid index that genuinely contains 1299 vectors** — it loads cleanly and passes every check. It is not a truncated prefix of the 8.08 MB one. Separately confirmed by experiment: truncating a healthy 8,084,693-byte index to 99.9 %, 95 %, 75 %, 53.8 % and 25 % makes it fail to load *every* time, so a torn file makes the index **vanish**, it does not shrink.

## Root cause — lost update, and it is structural rather than a race

- `load_vector_indexes()` is called from exactly two places, `db.rs:106` and `db.rs:169`, both inside `GraphDb::open()`. Nothing reloads the index afterwards — no refresh, no mtime check, no revalidation.
- Every write path (`db.rs:994`, `db.rs:1411`, `db.rs:1583`, plus the Node binding) takes the `Arc<RwLock<VectorIndex>>` from that open-time map, inserts, and immediately calls `idx.save(...)`.
- `SparrowDB.open()` takes no cross-process lock, so a second concurrent open succeeds.

So `save()` serialises whatever the handle has held since `open()` over whatever else has landed on disk meanwhile. No timing window is required: a long-running daemon that opened *before* a backfill reverts the entire backfill the next time it writes a single vector.

### 1. No lost-update detection (`vector_index.rs`)
As above. The file carries no generation, LSN or identity, so `save()` cannot tell "this is my next version" from "this would erase 1134 of someone else's vectors".

### 2. `save()` is not atomic — `vector_index.rs:178`
```rust
let bytes = bincode::serialize(self).map_err(std::io::Error::other)?;
std::fs::write(&path, bytes)?;
```
`fs::write` opens the destination with `O_TRUNC`: the good copy is destroyed before the replacement exists. No temp file, no fsync, no rename, no directory fsync.

The failure mode is worse than a shrink — it is total index loss. A torn file fails `bincode::deserialize`, `load_vector_indexes` swallows the error (`helpers.rs:368` is `if let Ok(Some(idx))`), the index disappears from the map, `get_vector_index` returns `None`, and every subsequent vector write is silently dropped. If the consumer then calls `createVectorIndex` again, a fresh **empty** index is created and saved straight over the damaged bytes.

There is also no integrity check at all: no length, no checksum. Measured on `main`, a shorter index image written over the front of a longer file (what an in-place overwrite or two interleaved writers produce) loads **silently as 5 vectors when 20 were written**, because `bincode::deserialize` ignores trailing bytes.

### 3. Re-insert is a silent no-op — `vector_index.rs:374`
```rust
if self.id_to_slot.contains_key(&node_id) {
    // Already present — skip (tombstone / update support is a future enhancement).
    return;
}
```
`insert` returns `()`, so re-embedding a node writes nothing and reports nothing. Backfill counts overstate what changed.

### 4. `addToVectorIndex` writes no node property (`sparrowdb-node/src/lib.rs`)
The KMSmcp backfill used exactly this API (`backfill-hnsw-embeddings.ts:29`), so 1721 vectors were written only into the HNSW file and are unrecoverable from column data — their store has 2591 `Knowledge` nodes and only 806 carrying `metadata.embedder_id`. Index and store diverge permanently and rebuild-from-columns is impossible.

### 5. `hybrid_search` reloads the index from disk on every call — `engine/expr.rs:217`
`VectorIndex::load(...)` per evaluation, ignoring the handle already held in `GraphDbInner`. `hybrid_search` is a scalar function evaluated per row, so a query over N rows performs N full reads and `bincode` decodes of the index — 8 MB of I/O and a complete graph rebuild per row for this consumer.

### 6. No load-time structural validation
An index whose `id_to_slot` disagrees with `nodes` silently rejects inserts for ids it believes it already holds while never returning them from a search, and the condition persists across restarts because it lives in the file.

## Follow-up outside this fix

`helpers.rs:368` swallows every `VectorIndex::load` error (`if let Ok(Some(idx))`). That is the amplifier that turns "one damaged file" into "the index does not exist and all writes are silently dropped". It should propagate or at minimum log loudly. Filed here for whoever owns `crates/sparrowdb/src/helpers.rs`.

ryaker:
Closes #441.

A production consumer lost 1134 embeddings with no error raised. The 4,352,529-byte survivor is a **complete, valid index holding 1299 vectors** — not a truncated prefix of the 8,084,693-byte / 2433-vector file it replaced. Truncation experiments confirm a torn HNSW file makes the index *vanish* (it fails to load and the engine then reports no index exists), so the shrink had a different cause.

That cause is a **lost update, and it is structural rather than a race**. `load_vector_indexes()` runs only inside `GraphDb::open()` (`db.rs:106`, `db.rs:169`); nothing reloads afterwards. Every writer mutates and re-saves the open-time snapshot, and `open()` takes no cross-process lock. A daemon that opened before a backfill reverts the entire backfill the next time it writes a single vector.

### 1. Lost-update detection — the floor, and it lands
Index files carry a generation counter. `save()` re-reads it immediately before writing and refuses when it disagrees with the generation this handle loaded, with an error beginning `"HNSW index generation conflict"` (`VectorIndex::is_lost_update`). The caller's write transaction aborts rather than committing a half-truth.

This also closes the amplification path: a fresh empty index can no longer overwrite a populated file, which is what the "load failed, so `createVectorIndex` makes a new one" sequence used to do.

Refusing is not the whole answer — a caller still has to decide whether to reload and retry — but a loud failure beats silently discarding another writer's vectors. Reload-and-merge is a reasonable follow-up; it needs a policy decision about which vector wins per node id, and it belongs with whoever owns the `db.rs` write paths.

### 2. Atomic save
Serialise → write `<path>.tmp` → `fsync` → `rename` → `fsync` the directory. A failed or interrupted save leaves the previous index intact.

Files gained a 36-byte header (magic, version, payload length, generation, CRC32C over header *and* payload). Damaged files are detected and **quarantined** as `<path>.corrupt.<millis>` rather than served or overwritten — the damaged bytes are the only surviving copy of the vectors, and a caller that reads a load failure as "no index here" would otherwise let the next save replace them.

Measured on the pre-fix tree: a shorter index image written over the front of a longer file (an in-place overwrite, or two writers interleaving) loads **silently as 5 vectors when 20 were written**, because `bincode` ignores trailing bytes. That path is now rejected.

v1 headerless files still load — existing deployments cannot be asked to re-embed — and are upgraded to v2 on the next save. v1 files also get a length cross-check against `bincode::serialized_size`, which catches the same trailing-bytes damage.

### 3. `insert` really updates, and says what it did
`insert` now returns `InsertOutcome::{Inserted, Updated}` and performs a genuine update: the vector is replaced in place, keeping the node's slot, id mapping and layer, and its outgoing links are rebuilt against the new position. Inbound links from nodes near the *old* vector are deliberately kept — they are sub-optimal navigation edges, never incorrect ones, and keeping them preserves reachability.

### 4. `addToVectorIndex` writes the node property
The consumer's backfill used exactly this API, so 1721 vectors were written only into the HNSW file and are unrecoverable from column data. It now writes `n.<property>` inside the same write transaction as the index update, so the embedding survives in the checkpointed column store.

Encoded as `sparrowdb:vec32:<little-endian f32 hex>`. The storage layer has no vector type — its `Value` is `Int64`/`Bytes`/`Float` — and the read path decodes `Bytes` with `String::from_utf8_lossy`, which mangles binary irreversibly. Hex keeps the value valid UTF-8 so it round-trips through the existing read path. Vectors above ~8189 dimensions cannot fit the overflow heap's 16-bit length field and are refused rather than left unrecoverable.

### 5. Load-time structural validation
`id_to_slot` count and consistency against `nodes`, slot ranges, entry-point layer, self-loops. Such an index silently rejects inserts for ids it believes it holds while never returning them from a search, and the condition persists across restarts because it lives in the file. Unlike a checksum failure this is *not* quarantined — the bytes decoded cleanly and may be recoverable by hand.

### 6. `hybrid_search` stops reloading per row
It is a scalar function evaluated per candidate row, and it called `VectorIndex::load()` every time — 8 MB of I/O and a full graph rebuild per row for this consumer. It now reuses a deserialised copy, revalidated against the file's generation + CRC read from the 36-byte header (no payload I/O), so staleness is impossible and the cost is one decode per index version.

Note: the in-memory `Arc<RwLock<VectorIndex>>` in `GraphDbInner` is not reachable from the execution layer without changing `Engine::new` in `engine/mod.rs`, which is outside this PR's file scope. The fingerprint-validated cache gives the same freshness guarantee, because every writer persists inside the same write lock that mutates.

## Tests

10 integration tests (`crates/sparrowdb/tests/vector_index_durability.rs`) and 14 storage unit tests. Every expected value is derived by hand from the fixture in a comment; nothing is a recording of program output.

**Measured against the pre-fix tree: 8 of the 10 integration tests fail.** The two that pass are stated as non-discriminating in their doc comments: `truncated_index_file_never_loads_as_fewer_vectors` (bincode already errored on truncation — it guards the new format, not the old bug) and `vectors_are_reachable_after_checkpoint_and_reopen` (a baseline round-trip guard).

Sample pre-fix failures:
```
a_stale_handle_cannot_silently_revert_another_writers_vectors:
  the stale handle's write must be refused, not silently applied
shorter_index_written_over_longer_file_is_rejected_not_silently_truncated:
  load() accepted a partially-overwritten index and returned 5 vectors;
  20 were written and 5 came back
failed_save_leaves_the_previous_index_intact:
  a save that cannot stage its bytes must fail loudly, not overwrite the good index
updated_vector_survives_save_and_reload:
  the reloaded vector must be the updated one (score 1.0), got 0
```

## Known gap, deliberately out of scope

`helpers.rs:368` is `if let Ok(Some(idx)) = VectorIndex::load(...)`, so it swallows every load error. That is the amplifier that turns one damaged file into "the index does not exist and all writes are silently dropped". Quarantine and the generation check contain the damage, but the error should propagate or at minimum be logged. `crates/sparrowdb/src/helpers.rs` is outside this PR's ownership; flagged in #441 for whoever owns it.

## Verification

- `cargo test -p sparrowdb-storage --no-fail-fast` — pass
- `cargo test -p sparrowdb-node --no-fail-fast` — pass
- `cargo test -p sparrowdb-execution --no-fail-fast` — pass
- `cargo test -p sparrowdb --no-fail-fast` — pass
- `cargo fmt --all -- --check` — clean, re-verified on a fresh clone of the pushed branch
- `cargo clippy --workspace --exclude sparrowdb-ruby --all-targets --keep-going -- -D warnings` — clean

## Summary by CodeRabbit

* **New Features**
  * Improved hybrid-search performance by reusing unchanged vector indexes.
  * Vector entries now persist with their associated node data and support safe replacement.
  * Added integrity checks and recovery safeguards for vector-index files.
  * Added support for upgrading legacy vector indexes to the latest format.

* **Bug Fixes**
  * Prevented corrupted, incomplete, or stale index data from causing invalid search results.
  * Reduced the risk of lost updates when multiple writers modify an index.

### DIFF
--- a/crates/sparrowdb-execution/src/engine/expr.rs
+++ b/crates/sparrowdb-execution/src/engine/expr.rs
@@ -147,6 +147,67 @@ impl Engine {
 
     // ── Hybrid search (issue #396) ────────────────────────────────────────────
 
+    /// Return the HNSW index for `(label, prop)` under `vec_dir`, reusing an
+    /// already-deserialised copy whenever the file on disk has not changed.
+    ///
+    /// `hybrid_search` is a scalar function: it is evaluated once per candidate
+    /// row.  The previous implementation called `VectorIndex::load()` on every
+    /// evaluation, so a query over N rows performed N full reads and `bincode`
+    /// decodes of the index file — for KMSmcp's 8 MB `Knowledge.embedding`
+    /// index that is 8 MB of I/O and a complete graph rebuild *per row*.
+    ///
+    /// Staleness is handled by validating the cheap on-disk fingerprint
+    /// (generation + CRC32C, read from the file's 36-byte header — no payload
+    /// I/O) before each reuse.  Because every writer persists the index in the
+    /// same write lock that mutates it, a matching fingerprint means the cached
+    /// copy is byte-identical to what the current writer would hand us.
+    ///
+    /// Returns `None` when the index does not exist or fails to load; the
+    /// caller then falls back to full-text-only results, exactly as before.
+    fn hybrid_vector_index(
+        vec_dir: &std::path::Path,
+        label: &str,
+        prop: &str,
+    ) -> Option<std::sync::Arc<sparrowdb_storage::vector_index::VectorIndex>> {
+        type Key = (std::path::PathBuf, String, String);
+        type Entry = (
+            (u64, u64),
+            std::sync::Arc<sparrowdb_storage::vector_index::VectorIndex>,
+        );
+        static CACHE: std::sync::OnceLock<std::sync::Mutex<HashMap<Key, Entry>>> =
+            std::sync::OnceLock::new();
+
+        // Missing file / unreadable header → nothing to search.
+        let fingerprint =
+            sparrowdb_storage::vector_index::VectorIndex::fingerprint(vec_dir, label, prop)
+                .ok()
+                .flatten()?;
+
+        let cache = CACHE.get_or_init(|| std::sync::Mutex::new(HashMap::new()));
+        let mut guard = cache.lock().ok()?;
+        let key: Key = (vec_dir.to_path_buf(), label.to_owned(), prop.to_owned());
+
+        if let Some((cached_fp, idx)) = guard.get(&key) {
+            if *cached_fp == fingerprint {
+                return Some(std::sync::Arc::clone(idx));
+            }
+        }
+
+        let loaded = sparrowdb_storage::vector_index::VectorIndex::load(vec_dir, label, prop)
+            .ok()
+            .flatten()?;
+        let idx = std::sync::Arc::new(loaded);
+
+        // Bound the cache: a process that opens many databases should not
+        // retain every index it has ever queried.
+        const MAX_CACHED_INDEXES: usize = 8;
+        if guard.len() >= MAX_CACHED_INDEXES && !guard.contains_key(&key) {
+            guard.clear();
+        }
+        guard.insert(key, (fingerprint, std::sync::Arc::clone(&idx)));
+        Some(idx)
+    }
+
     /// Evaluate `hybrid_search(label, emb_prop, text_prop, query_vec, query_text, k[, alpha])`.
     ///
     /// Runs vector search (HNSW) + BM25 full-text search independently, then
@@ -214,10 +275,10 @@ impl Engine {
         // ── 1. Vector search ────────────────────────────────────────────────
         let vec_dir = self.snapshot.db_root.join("vector_indexes");
         let vec_results: Vec<(u64, f32)> =
-            match sparrowdb_storage::vector_index::VectorIndex::load(&vec_dir, &label, &emb_prop) {
+            match Self::hybrid_vector_index(&vec_dir, &label, &emb_prop) {
                 // ef = max(k*2, 50) gives a reasonable exploration budget.
-                Ok(Some(idx)) => idx.search(&query_vec, k * 2, (k * 2).max(50)),
-                _ => vec![],
+                Some(idx) => idx.search(&query_vec, k * 2, (k * 2).max(50)),
+                None => vec![],
             };
 
         // ── 2. Full-text (BM25) search ───────────────────────────────────────
--- a/crates/sparrowdb-node/src/lib.rs
+++ b/crates/sparrowdb-node/src/lib.rs
@@ -65,6 +65,67 @@ fn validate_cypher_identifier(s: &str) -> napi::Result<()> {
     }
 }
 
+/// Prefix identifying a node property that holds an embedding written by
+/// `addToVectorIndex`.  Self-describing so a recovery tool can find and decode
+/// these without guessing.
+const VEC32_PREFIX: &str = "sparrowdb:vec32:";
+
+/// Encode `vector` as a node property value: `sparrowdb:vec32:<hex>`, where
+/// `<hex>` is the little-endian IEEE-754 bytes of each `f32` in order.
+///
+/// # Why a text encoding
+///
+/// The storage layer's `Value` has three variants — `Int64`, `Bytes`, `Float` —
+/// and no vector type, so a vector reaching the property path today is coerced
+/// to `Int64(0)`: `SET n.embedding = $vec` stores a zero and puts the real
+/// embedding only in the HNSW file.  `Bytes` *can* carry the data (values over
+/// 7 bytes spill to the overflow heap), but the read path decodes `Bytes` with
+/// `String::from_utf8_lossy`, which mangles arbitrary binary irreversibly.
+/// Hex keeps the value valid UTF-8, so it round-trips through the existing
+/// read path unchanged and stays decodable by `decode_vector_property`.
+///
+/// Returns `None` when the encoded form would exceed the 65535-byte limit of
+/// the overflow heap's length field (about 8189 dimensions).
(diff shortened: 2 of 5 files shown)


ryaker:
**On splitting this PR** — the coordinator offered to split atomicity + the property fix from generation-stamped concurrency detection. I kept it as one PR.

Reason: the six changes are one causal chain, not six independent fixes, and splitting them leaves a window where the halves actively undermine each other. The generation check is what stops a fresh empty index from overwriting a populated file; without it, quarantine only preserves the damaged bytes under a `.corrupt` name while the canonical path still gets an empty index written over it. Conversely, the generation counter has to live somewhere, and the only place to put it is the header that atomicity introduced — landing them separately means writing the file format twice and shipping an intermediate on-disk version nobody will ever run.

The whole change is also confined to three files plus one new test file, and every change is covered by tests that were measured to fail against the pre-fix tree. If review disagrees, the natural cut is (1,2,5) file-format + durability against (3,4,6) API behaviour, and I can resplit along that line.

**One correction to the incident analysis.** The refutation of the torn-write hypothesis was based on truncation experiments, and those are correct as far as they go — bincode rejects every short read. But truncation is not the only damage a non-atomic write produces. Measured on the pre-fix tree: a *shorter* index image written over the front of a longer file loads **silently as 5 vectors when 20 were written**, because `bincode::deserialize` ignores trailing bytes. That produces exactly the reported signature — a complete, valid, silently smaller index — and truncation-only experiments do not cover it.

I am not claiming this is what happened. The lost-update mechanism is structurally proven and fits the numbers, and it is fixed here as the top item. But the trailing-bytes hole is real, was silent, and is now closed, so it is worth recording that the two hypotheses were not mutually exclusive.

PR #447, opened 2026-08-02
fix(vector): guarantee HNSW inserts stay reachable (closes #443)

### LINKED ISSUE #443: HNSW: ~60% of novel embeddings are stored but unreachable by vectorSearch; no API to distinguish stored-vs-findable
## Summary

Vectors inserted into the HNSW index are stored correctly but **cannot be found by `vectorSearch`**. Measured against a real production store (KMSmcp) with genuine Ollama `nomic-embed-text` embeddings of novel text: **2 of 5 findable — roughly a 60% recall loss on new writes.** Silent; nothing surfaces it.

## Measurement, and why earlier numbers were wrong

Three probes give three different answers, and only one is representative:

| Probe | Findable |
|---|---|
| Random orthogonal 768-d unit vector | ~0/5 (score 0.05–0.12) |
| Near-duplicate of an already-reachable vector | 5/5 (score ~1.0) |
| **Real embedding of novel text** | **2/5** |

The first is adversarial — independent random unit vectors in 768-d are near-orthogonal to everything including greedy search's own descent path. The second is trivially easy — it lands beside a node the graph already reaches. The third is what production does.

Counts on the live store confirm nodes are present but unreached:

```
graph nodes            = 2595
id_to_slot entries     = 1347
vectorSearch reachable = 1306      <- 41 stored but unreachable
```

`reachable` moved 1306 → 1308 for the 2 that took, so the other 3 are in `nodes[]` and simply never visited.

## Ruled out — do not re-investigate

- **Structural corruption.** `nodes.len()` and `id_to_slot.len()` agree exactly across three snapshots (2489/2489, 1340/1340, 1347/1347). `slot_mismatch=0, missing_from_map=0, isolated=0, bad_dim=0`, entry point valid.
- **Degenerate level generator / frozen skeleton.** Layer histogram is healthy — 94.4% / 5.3% / 0.3% across layers 0/1/2, matching geometric decay for `m=16`, and layer ≥1 nodes span the full id range to 2596. `entry_point` being pinned at 227 is a consequence, not a cause. (Hypothesised then withdrawn on the data by the reporting session.)
- **Insert failing.** The index grew 1340 → 1347 during investigation. Inserts land.

So this is a **graph-connectivity / search-quality** problem, not a storage or structure problem: newly appended nodes get connection lists but greedy descent from the entry point does not reach them.

Separately: ~40–56 nodes are *stably* unreachable across all snapshots — the same low-id set each time, not growing, not correlated with recent inserts. Plausibly ordinary HNSW recall on nodes inserted when the graph was tiny. Lower severity, likely a different cause.

## The API gap that made this hard to see

**"Not returned by `vectorSearch`" does not mean "absent from the index"** — and there is no way to ask the right question. `insert()` returns `()`, so a caller cannot distinguish:

- inserted
- skipped because already present (`vector_index.rs:374` early-return)
- stored but unreachable

All three look identical. Selecting insert targets by subtracting `vectorSearch` results from `id(k)` therefore picks nodes that *already* occupy a slot, whose inserts then silently no-op — which produced a false "writes are entirely broken" conclusion during investigation, corrected only by dumping `id_to_slot` from the bincode payload directly.

**Requested:** a `has_vector(label, prop, node_id)` accessor, and/or `insert()` returning its outcome. PR #442 already changes the Rust `insert()` to return `InsertOutcome::{Inserted,Updated}`, but the Node binding still returns `void` and there is still no way to ask whether a node has a vector.

## Priority

Per the reporting session: this is the most damaging of the three vector defects because it is silent and compounds as the corpus grows, though the lost update (#441) is worse per-incident since it destroyed ~1150 vectors in one event.

## Related

- #441 / #442 — lost update, atomicity, `addToVectorIndex` not writing the node property
- `helpers.rs:368` swallows index load errors, flagged in #441

ryaker:
Closes #443.

Based on `fix/vector-index-durability` (#442) — review that first; this PR is the diff on top of it.

## Root cause — confirmed, and it is the reciprocal-link theory

`insert` wires `new -> neighbours` unconditionally. The reciprocal step, `neighbours -> new`, only lands if the new node beats an existing occupant of that neighbour's list on distance:

```rust
if nb_connections.len() < m_max { push(slot) }
else { /* keep the m_max closest of conns ∪ {slot} */ }
```

Lose all 32 contests — which is what happens in a saturated neighbourhood — and the node has **zero incoming edges**. Greedy descent only follows *outgoing* edges from `entry_point`, so it is invisible to `search` at any `k` and any `ef`. Stored, reported present by every available API, never returned.

The same eviction strands **existing** nodes: the prune drops an edge, and if it was the last edge into that node, it is gone from the reachable set permanently. Nothing repaired it — re-inserting takes the "already present" path, so not even a full backfill could heal it.

Live store: **1347 stored, 1307 reachable, 40 stranded** — 33 pure sinks and 7 forming a closed island with non-zero in-degree and no path back.

**Reachability, not in-degree, is the property that matters.** An in-degree check calls those 7 healthy. Every assertion here uses a traversal.

## Reproduced locally

Neither of the two probes in the issue reproduces this. Random orthogonal vectors are near-orthogonal to everything; near-duplicates land beside a node the graph already reaches. What reproduces it is **anisotropic** geometry — real sentence embeddings sit in a narrow cone, not spread over the sphere, so pairwise distances are compressed and a new node must squeeze into lists whose occupants are all at similar range — plus **cluster-by-cluster arrival** rather than a shuffled stream.

| fixture | pre-fix unreachable | post-fix |
|---|---|---|
| 700 vectors, 48-d, aniso 2.0, batched (the test fixture) | 2 (both zero in-degree) | **0** |
| 1400 vectors, 128-d, aniso 4.0 | 14 | **0** |
| 1400 vectors, 128-d, aniso 2.0, batched | 7 | **0** |
| 1300 clustered + 100 novel outliers | 74 | **0** |

## The fix — one invariant at every edge eviction

> A node that displaces an edge `nb -> x` must itself point at `x`.

Then any path through `nb -> x` reroutes as `nb -> slot -> x`; `slot` is reachable because `nb` is; the reachable set can never shrink. Three parts:

- **`link_back`** replaces the inline prune. It keeps the original quality rule — a candidate is admitted only if it beats the furthest occupant — but adopts whatever it displaces *when that edge was the last one in*. `has_other_inbound` makes adoption conditional, so the common harmless eviction costs nothing and graph quality is untouched. (An unconditional version cost 7 points of recall on clean data; that is why it is conditional.)
- **A last-resort forced link** when a new node loses every contest, so no insert can end at zero in-degree.
- **`adopted`** guards the second-order failure that survived my first draft: adopted nodes are by construction far from `slot`, so without the set the *next* adoption evicts the previous one. Concretely, inserting node 368 removed the last inbound edge of node 194. Worth knowing about if you review this logic.

**`repair()`** reconnects anything already stranded — necessary regardless, because the insert-time guarantee does nothing for an index written by an older build, and existing stores do not self-heal.

## Recall — measured, not assumed

recall@10 vs brute force on a **held-out** query set, 10 fixtures:

| | pre-fix | post-fix |
|---|---|---|
| unchanged | 8 fixtures | — |
| 128-d aniso 4.0 | 0.7042 | 0.7025 (−0.002) |
| 1300 + 100 outliers | 0.5983 | 0.6808 (+0.083) |
| **mean** | **0.8518** | **0.8598** |

Connectivity is not bought with recall.

## Observability — the reason this survived months

`vectorSearch` answers a different question than "is this vector stored", and there was no way to ask the right one. Three states looked identical: inserted / skipped-as-present / stored-but-unreachable.

- `VectorIndex::has_vector`, `reachable_count`, `unreachable_ids`, `repair`
- Node: `hasVector`, `vectorIndexHealth` → `{stored, reachable, unreachable}`, `repairVectorIndex`

`stored != reachable` is the single number that would have caught this.

## Tests

`crates/sparrowdb/tests/regression_443.rs` (7) + 6 unit tests in `vector_index.rs` reaching private fields for in-degree and capacity. Every expected value is hand-derived in a comment; nothing is a recording of program output.

**Pre-fix failure proof** — measured by reverting only `link_slot` and keeping the new accessors so the tests still compile. **6 of 7 fail** with the reported symptom:

| test | pre-fix result |
|---|---|
| `spa443_no_stored_vector_is_unreachable` | 2 of 700 unreachable: `[234, 244]` |
| `spa443_every_vector_retrieves_itself_given_full_budget` | 1 of 700 missing at `ef = N`: `[244]` |
| `spa443_reachability_survives_save_and_reload` | reachable 498, expected 500 |
| `spa443_has_vector_reports_storage_independently_of_search` | reachable 698, expected 700 |
| `spa443_repair_is_safe_to_call_on_a_healthy_index` | `repair()` returned 1, expected 0 |
| `spa443_updates_do_not_strand_vectors` | 1 stranded by in-place update: `[344]` |
| `spa443_every_vector_retrieves_itself_at_default_ef` | **passes pre-fix** |
| `spa443_every_stored_vector_is_reachable` (unit) | 2 of 700 unreachable |
| `spa443_no_node_has_zero_in_degree` (unit) | 2 pure sinks: `[234, 244]` |

The one that passes pre-fix is labelled **in-file** as a forward guard, not presented as evidence — `regression_406.rs` is the cautionary example.

Includes a save/reload case, and inserts *after* reload, since a class of defect here only shows up on the far side of persistence.

## Known limitation — not in scope, should be filed

Reachability is necessary for retrieval but **not sufficient**. `select_neighbours` takes the nearest `m_max` with no diversity heuristic (HNSW Algorithm 4), so the graph carries few long-range links. A strong outlier reachable through a single inbound edge can still be missed at `ef = 50` because greedy descent never brings the edge's holder into the candidate window. Measured on the harsher fixture: **2 of 700**, in-degree 1, retrievable at `ef = 800` but not at `ef ≤ 200`.

This is present before and after this change and is unaffected by it. I prototyped Algorithm 4 with `keepPrunedConnections` and A/B'd it: **+0.01–0.02 recall, and it did not close the gap** — so it is not the lever, and swapping neighbour selection is a design change that needs its own recall validation. Filing separately rather than folding it in here.

## Note on the issue's framing

The "~60% of novel embeddings" headline in #443 is a sampling artefact (five consecutive internal ids competing in one saturated neighbourhood). The real rate is **1–5%**. The severity argument is unchanged and arguably stronger: nothing repaired it, so the loss was unbounded over time.

## Verification

- `cargo test -p sparrowdb-storage --no-fail-fast` — 167 passed, 0 failed
- `cargo test -p sparrowdb --test regression_443 --no-fail-fast` — 7 passed
- `cargo fmt --all -- --check` — clean, re-verified in a fresh clone of the pushed branch
- `cargo clippy --workspace --exclude sparrowdb-ruby --all-targets --keep-going -- -D warnings` — clean

### DIFF
--- a/crates/sparrowdb-node/src/lib.rs
+++ b/crates/sparrowdb-node/src/lib.rs
@@ -261,14 +261,15 @@ pub struct NodeResult {
 ///
 /// ```typescript
 /// interface VectorIndexHealth {
-///   /** Vectors stored in the index. */
-///   stored: number;
-///   /** Vectors `vectorSearch` can actually reach. */
-///   reachable: number;
-///   /** `stored - reachable`; must be 0 on a healthy index. */
-///   unreachable: number;
+///   stored: number;       // vectors stored in the index
+///   reachable: number;    // vectors `vectorSearch` can actually reach
+///   unreachable: number;  // stored - reachable; must be 0 when healthy
 /// }
 /// ```
+///
+/// Note the `//` comments above rather than nested doc comments: a `*/` inside
+/// a generated JSDoc block closes it early, which is why the checked-in
+/// `npm/sparrowdb/index.d.ts` does not currently parse as TypeScript.
 #[napi(object)]
 pub struct VectorIndexHealth {
     /// Vectors stored in the index.
--- a/crates/sparrowdb-node/src/lib.rs
+++ b/crates/sparrowdb-node/src/lib.rs
@@ -255,6 +255,30 @@ pub struct NodeResult {
     pub score: f64,
 }
 
+// ── VectorIndexHealth ─────────────────────────────────────────────────────────
+
+/// Coverage report for an HNSW index — see `SparrowDB.vectorIndexHealth`.
+///
+/// ```typescript
+/// interface VectorIndexHealth {
+///   /** Vectors stored in the index. */
+///   stored: number;
+///   /** Vectors `vectorSearch` can actually reach. */
+///   reachable: number;
+///   /** `stored - reachable`; must be 0 on a healthy index. */
+///   unreachable: number;
+/// }
+/// ```
+#[napi(object)]
+pub struct VectorIndexHealth {
+    /// Vectors stored in the index.
+    pub stored: u32,
+    /// Vectors reachable by greedy traversal from the entry point.
+    pub reachable: u32,
+    /// `stored - reachable`. Non-zero means silent recall loss.
+    pub unreachable: u32,
+}
+
 // ── SparrowDB ─────────────────────────────────────────────────────────────────
 
 /// Top-level database handle.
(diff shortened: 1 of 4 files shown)


ryaker:
Follow-ups filed, as promised in the description:

- **#448** — `select_neighbours` has no diversity heuristic; reachable outliers still unretrievable at default `ef`. This is the navigability half that this PR deliberately does not touch, with the Algorithm 4 A/B results recorded so they are not re-derived.
- **#449** — `npm/sparrowdb/index.d.ts` does not parse as TypeScript (144 errors at HEAD). Found while adding the bindings here, not caused by them; the second commit on this branch avoids adding to it and incidentally drops the count to 35.

ryaker:
@coderabbitai review

Issue #451, opened 2026-08-02
Fail-closed open() is one-shot: after quarantine, the store reopens 'healthy' and silently drops vector writes again


ryaker:
## The residual hole after #441/#442/#445/#446/#447

All four merged, and the original silent-write-drop can still recur. The sequence:

1. Index becomes damaged.
2. **First** `open()` → loud `Err(Corruption)`, and the damaged bytes are quarantined to `<path>.corrupt.<millis>`. Correct.
3. **Second** `open()` → the `.bin` no longer exists, so this is now the *absent* case. `open()` succeeds. No index is registered for that `(label, prop)`.
4. Every subsequent vector write is silently dropped, and every search returns nothing — **exactly the behaviour #446 was written to eliminate.**

The condition remains *detectable* — `vector_index_load_failures()` reports the quarantine artifact — but nothing forces anyone to look. A daemon restarting under `KeepAlive` sails straight past step 3 into a healthy-looking store that is quietly discarding data.

## Why the obvious fix is wrong

Making `open()` fail on the presence of a quarantine artifact was considered and rejected, and I agree with the reasoning: once bytes are out of service, refusing to start on every subsequent boot forces an operator to choose between "the store won't start" and "delete the only surviving copy of my vectors." That is a worse trade than the one it prevents.

## The durable fix belongs on the write path

The real defect is that **writing a vector to a `(label, prop)` with no registered index is a silent no-op.** That is the amplifier. If the write path refused — or at minimum surfaced — a vector write against a missing index, none of the above would be silent, and the same guard would also catch the plain misconfiguration case where someone writes embeddings before calling `createVectorIndex`.

Options, roughly in order of preference:

1. **Error on vector write with no index for that `(label, prop)`.** Loud, and it cannot be missed. Risk: breaks callers who currently write embeddings to un-indexed properties on purpose and expect them stored only as node properties.
2. **Auto-register a fresh index** when a quarantine artifact exists for the pair, so writes land somewhere recoverable and coverage rebuilds from that point forward.
3. **Surface it in `vectorIndexHealth()`** as a distinct state — `quarantined_present_no_live_index` — so a health check catches it even though `open()` succeeded.

(1) and (3) compose well. (2) is attractive for KMSmcp specifically, since their vectors written via `addToVectorIndex` are unrecoverable and continuing to drop them makes the hole deeper.

## Scope note

[...]

Issue #450, opened 2026-08-02
#442 + #446 do not compose: quarantine renames the file, so vector_index_load_failures() reports [] when open() fails


ryaker:
## Found by composing the two branches

`fix/vector-index-durability` (#442) and `fix/index-load-errors-surfaced` (#446) each pass their own suites. Merged together, `vector_index_load_errors::absent_index_and_unloadable_index_are_distinguishable` **fails**.

They merge without conflict — this is a behavioural interaction, not a textual one, so nothing short of building and running the combination catches it.

## The interaction

Measured on the merged tree with a damaged index (`hnsw_Memory_embedding.bin` truncated to 4 bytes):

```
BEFORE:  ["hnsw_Memory_embedding.bin"]
open()   -> Err(Corruption)   ← correct, and the message is excellent
AFTER:   ["hnsw_Memory_embedding.bin.corrupt.1785645973245"]
vector_index_load_failures() -> []          ← wrong
```

#442 **quarantines** the damaged file by renaming it to `.corrupt.<millis>`. #446's `vector_index_load_failures()` scans for `*.bin` and filters out quarantine artifacts by design — correctly, in isolation. Composed, the `.bin` no longer exists by the time the accessor runs, so it reports nothing.

## Why it matters

`vector_index_load_failures()` exists specifically to be callable on a database that **refuses to open** — it is the diagnostic path, and it never returns `Err`. Returning `[]` is indistinguishable from a healthy store. A consumer health check built on it would report "no problems" on exactly the database that will not start.

That is the same defect class this whole series has been fixing: a value that reads as fine when it is not.

Mitigating: the `open()` error itself is self-contained and genuinely good — it names the label/prop, the original path, the decode failure, the quarantine path it preserved the bytes to, why it refused, and the recovery step. So the information is not *lost*, only absent from the accessor.

## Suggested fix

Have `vector_index_load_failures()` also enumerate `*.bin.corrupt.*` artifacts and report them, with the quarantine path and the reason recorded at quarantine time. That preserves #446's intent (quarantine artifacts must never be *loaded* as live indexes) while making them *visible* to the diagnostic.

Alternative — report before quarantining — is worse: it would either duplicate the scan or make the accessor stateful.

## Repro

```
git checkout -b integ origin/fix/issue-443-hnsw-reachability   # contains #442
git merge origin/fix/index-load-errors-surfaced                 # #446, no conflicts
cargo test -p sparrowdb --test vector_index_load_errors
[...]

Issue #449, opened 2026-08-02
npm/sparrowdb/index.d.ts does not parse as TypeScript (144 syntax errors at HEAD)


ryaker:
Found while adding bindings in #447. Not caused by it.

## Problem

The checked-in, published type declarations do not parse. `tsc --noEmit --skipLibCheck` against `npm/sparrowdb/index.d.ts` at `main` reports **144 syntax errors**, the first at line 32:

```
index.d.ts(32,2): error TS1109: Expression expected.
index.d.ts(32,8): error TS1005: ';' expected.
index.d.ts(33,2): error TS1109: Expression expected.
...
```

Reproduce:

```bash
npx typescript@5 tsc --noEmit --skipLibCheck npm/sparrowdb/index.d.ts
```

## Cause

Several Rust doc comments in `crates/sparrowdb-node/src/lib.rs` illustrate the generated TypeScript using *nested* doc comments. NAPI-RS copies them verbatim into a JSDoc block, and the inner `*/` closes the outer block early — everything after it is parsed as code. From the generated file:

```typescript
/**
 * A single result row from a vector or full-text search.
 *
 * ```typescript
 * interface NodeResult {
 *   /** Node ID as a decimal string (full u64 range, JS-safe). */   <-- closes the outer comment here
 *   id: string;
 *   ...
```

`NodeResult` is the first occurrence; there are more further down.

## Impact

Any TypeScript consumer importing `sparrowdb` gets a broken declaration file. It degrades silently — `skipLibCheck` is on by default in many setups, which suppresses *checking* the file but not the parse errors, and editors surface it as missing/incorrect types rather than as a broken package.

## Fix

In `crates/sparrowdb-node/src/lib.rs`, replace nested `/** ... */` inside the ```typescript examples with `//` line comments. #447 does this for the one struct it adds (`VectorIndexHealth`) and, as a side effect of re-syncing the comment nesting, brings the count from 144 to 35 — the remaining 35 are the other occurrences.

Worth adding a CI step so it cannot regress:

```bash
npx tsc --noEmit npm/sparrowdb/index.d.ts
```

Note the file is regenerated by `napi build`, so the fix belongs in the Rust doc comments, not in the `.d.ts`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Aq3C5J646sBCH7sVs3uaCE

Issue #445, opened 2026-08-02
Vector index loader swallows load errors: a damaged index file is indistinguishable from an absent one (silent total vector outage)


ryaker:
## Summary

`load_vector_indexes` in `crates/sparrowdb/src/helpers.rs` discards every error from
`VectorIndex::load`:

```rust
if let Ok(Some(idx)) = sparrowdb_storage::VectorIndex::load(&dir, label, prop) {
    map.insert((label.to_string(), prop.to_string()), Arc::new(RwLock::new(idx)));
}
```

A damaged, unreadable, or rejected index file does not surface as an error. It surfaces
as **"there is no index for (label, prop)"** — the exact same observable as a database
that never had an index configured at all. The function's own doc comment says
"Unreadable files are silently skipped", so the skip is deliberate; the consequence
downstream was not.

## Production consequence

Collapsing *absent* and *present-but-unloadable* into one outcome is the whole bug.
A missing key in the `vector_indexes` map is the signal every downstream path uses to
mean "no index configured", so one damaged file silently produces:

- every subsequent vector write for that `(label, prop)` is dropped, because the
  insert path only writes when the map contains the key;
- every `vectorSearch` returns nothing;
- the engine reports *"no vector index; call createVectorIndex first"*, which reads to
  an operator as a configuration mistake rather than as data loss.

This was observed in production: a store that "refuses inserts, persistently across
restarts". The real state was one index file the loader had quietly skipped. Nothing in
any log, error, or query result distinguished it from an unconfigured store.

## Why this blocks #442

PR #442 adds header/CRC validation and quarantines damaged index files to
`<path>.corrupt.<millis>` instead of serving them. That is correct on its own, but
combined with this swallow it converts a silent partial data loss into a **silent total
outage**: the file is quarantined, the loader finds nothing, and the consumer sees zero
vectors with no error at all. #442 cannot safely be deployed to the affected store until
the swallow is fixed.

## Why the property-index stance does not apply here

`property_index.rs` documents that a load failure is "not fatal; the next `Engine` will
simply rebuild from column files". That is right for the property index because it is
*derived* state — discarding it costs time, not data.

An HNSW index is not derived state. Vectors added through `addToVectorIndex` (#441) live
only in that file; there is no column data to rebuild them from. A load failure there is
data loss, and continuing to run compounds it by dropping every write that follows.

## Proposed fix

Fail-closed, with a diagnostic:

[...]

PR #458, opened 2026-08-02
fix(vector): make VectorIndex::load non-destructive (closes #456, closes #455)

### LINKED ISSUE #455: vector_index_load_failures() live-arm reports a path that no longer resolves (load() renames it during the same probe)
## Summary

`crates/sparrowdb/src/helpers.rs::vector_index_load_failures` has two arms:

- **live arm** — the `.bin` is present and fails to load
- **artifact arm** — a `.bin.corrupt.<millis>` quarantine file is present

The live arm reports the **`.bin` path**. But `VectorIndex::load` *quarantines by renaming that file during the same probe*, so by the time the caller receives the result, the path it names no longer resolves.

Effect: a consumer that does `fs.existsSync(entry.path)` — the obvious sanity check, and the one a health check would use to confirm damage is real — gets `false` on the very first report. On the **next** call the artifact arm fires and reports the quarantine path correctly.

## Why this is more than cosmetic

The whole point of the 4-tuple `(label, prop, path, reason)` is that the **path is the machine-readable part** — the `reason` is reconstructed, not the original error text, so it is explicitly not authoritative. If the path is transiently wrong, the only trustworthy field in the first report is the `(label, prop)` pair.

A consumer building an alert on "damage detected, here is the file" will emit a path that has already moved. Given #451 means the *first* boot after damage is the only one that errors loudly, the first report is exactly the one most likely to be acted on.

## Suggested fix

Have the live arm report the path the file was quarantined **to**, not the path it was found at. `VectorIndex::load` knows the destination — it constructs it. Threading it into the error, or having the loader hand back both paths, keeps the field honest.

Alternative: report both, as `path` and `quarantined_to`.

## Notes

- The bytes are correctly named in `reason`, so nothing is lost — this is about which field a caller can trust.
- Covered explicitly rather than papered over: test 4 in `npm/sparrowdb/test/integration.test.js` (added in #454) encodes the live-arm behaviour as-is, and test 5 asserts the artifact-arm path **does** resolve.
- Found while adding the napi binding (#454); out of that agent's file ownership.

## Related

- #446 — added `vector_index_load_failures`
- #451 — fail-closed `open()` is one-shot, which is why the first report matters most
- #454 — napi binding

### LINKED ISSUE #456: CRITICAL: #442 made VectorIndex::load() destructive — a hybrid_search READ quarantines the index and returns zero hits silently
## Live on main right now, unfixed, untracked

#442 moved the quarantine rename **inside** `VectorIndex::load()`:

```rust
// crates/sparrowdb-storage/src/vector_index.rs:607
name.push(format!(".corrupt.{stamp}"));
```

`load()` is now destructive, and **every existing caller inherited that**. Three non-test callers on `main`:

| Caller | Should it mutate? |
|---|---|
| `helpers.rs:453` `load_vector_indexes` — open path | **yes**, intended |
| `helpers.rs:519` `vector_index_load_failures` — the diagnostic | **no** |
| `crates/sparrowdb-execution/src/engine/expr.rs:196` — `hybrid_search` | **no — this is a query path** |

```rust
// expr.rs:196-198
let loaded = sparrowdb_storage::vector_index::VectorIndex::load(vec_dir, label, prop)
    .ok()
    .flatten()?;
```

## Effect: a read query silently destroys and lies

Reproduced by execution against the composed tree — open a db, create + populate + save an index, `hybrid_search` → 1 hit. Truncate the `.bin` to 4 bytes **without closing the db**, so open-time validation never sees it. Run `hybrid_search` again:

```
hybrid_search after corruption:  Ok, hit_count = 0
vector_indexes/ after query:     ["hnsw_Doc_embedding.bin.corrupt.<millis>"]
get_vector_index still Some:     true
```

A **read** returns `Ok` with zero hits, no error, no log line — and renames the last live copy of the index aside as a side effect.

The asymmetry matters: plain `vectorSearch` is unaffected, because it reads the in-memory `Arc` cached at open. Only `hybrid_search`'s independent disk read is exposed.

**#442 edited these exact lines** — its perf change replaced a per-row `VectorIndex::load()` with the cached loader, and the swallow survived the edit.

## Correction for anyone grepping

An earlier draft of this finding cited `expr.rs:215-220` with `match … { Ok(Some(idx)) => …, _ => vec![] }`. That is the **pre-#442** shape and no longer exists. The swallow moved to `.ok().flatten()?` at line 196. Grepping the old text will wrongly suggest it is fixed.

## Suggested fix — one change, not three

Split `load()` (non-destructive) from `load_and_quarantine()` (used **only** by the open path). That fixes the query path and the diagnostic together.

The diagnostic case interacts badly with #450 and #455: calling `vector_index_load_failures()` renames the `.bin`, then the next call reports differently. **A health check that runs twice gets a different answer than one that runs once, and the first run caused the change.**

## Also unfixed on main, from the same review

- **N damaged indexes require N failed restarts.** `load_vector_indexes` returns on the *first* `Err`, so only one bad file is probed (and quarantined) per `open()`. With two planted: restart 1 → `Err`, restart 2 → `Err`, restart 3 → `Ok` with **both** silently "no index configured" and writes for both dropped invisibly. #451 flags this for one file; it scales linearly and no test plants 2+.
- **Stale quarantine artifacts false-flag a healthy store forever.** Quarantine, then create a fresh healthy index for the same `(label, prop)`: `scan_vector_index_dir` reports both, so the diagnostic returns a failure for a completely healthy store. Mirror image of #450.
- **Dangling symlink classifies as absent.** Scan classifies by filename, so a symlink ending `.bin` reads as `Live`, but `load()` starts with `if !path.exists()`, which follows symlinks → false → `Ok(None)` → "absent". No error, no quarantine, invisible to the accessor. Exactly the present-but-treated-as-absent case #445 exists to kill. Narrow trigger; medium severity.

## Coverage of the audit that found this

~230 candidate swallow sites examined outside tests. ~200 safe by design (parsing fallbacks, WAL torn-record stop-on-first-error, derived-cache rebuilds, the deliberate 3-tier HWM recovery chain, lock-poison guards). 1 fixed by #446. **5 same-class unfixed. 2 arising from composition.**

## Provenance

Adversarial review by the KMSmcp consumer, demonstrated by execution. Reported to this session and **missed** — it arrived in a bus message that expired unread while the same review's other findings were being acted on.

ryaker:
Closes #456. Also closes #455 — reasoning below.

Rebased onto c9f7a0b, so this sits on top of both #453 (the `save()` race) and #454 (the napi binding). #454's binding is migrated here, and the composition with #453 is asserted by execution rather than by reading the two patches side by side — see test 9.

## The cost that applies even when nothing is broken

Every call to `GraphDb::vector_index_load_failures` fully deserialised every **healthy** index it found. For an 8 MB `Knowledge.embedding` index polled hourly, that is 8 MB of I/O and a complete HNSW graph rebuild an hour, spent confirming that nothing is wrong.

That is a permanent cost in the common case. The destructive read this issue is named for is a correctness bug in an edge case. Both argue for the same change — splitting judgement from side effect, and cheap from deep — but the cost argument stands even for a reader who thinks the mutation was tolerable.

## The correctness bug

#442 moved the quarantine rename **inside** `VectorIndex::load()`. `load()` became destructive and every caller inherited it. Three non-test callers; one was entitled to mutate.

| caller | may mutate? | what it got |
|---|---|---|
| `helpers.rs` `load_vector_indexes` — open | yes, intended | correct |
| `helpers.rs` `vector_index_load_failures` — diagnostic | **no** | mutated while diagnosing |
| `expr.rs:196` `hybrid_search` cached loader — query | **no** | `.ok().flatten()?` swallowed it |

Measured against the pre-fix tree (probe deleted before commit): open a db, create + populate + save an index, `hybrid_search` → `List([{node_id: 0, …}])`. Truncate the `.bin` to 4 bytes **without closing the db**. `hybrid_search` again:

```
AFTER: Ok, value = List([])          hit_count = 0
bin still present: false
artifacts: ["hnsw_Doc_embedding.bin.corrupt.1785700052598"]
```

A read returned success, returned nothing, and renamed the last live copy of the index aside. Same fixture on this branch:

```
AFTER: Ok, value = Null              (+ a tracing::warn naming file and reason)
bin still present: true
artifacts: []
```

Against over-claiming: the pre-fix mutation was **idempotent**. A given damaged file is quarantined once; repeat polls do no further I/O. It destroyed exactly once per incident — silently, from something shaped like a getter.

## split_design

**`sparrowdb-storage`** — judgement in one place, side effect opt-in:

```rust
fn read_index(path) -> io::Result<LoadOutcome>   // private; Absent | Loaded | Unreadable | Undecodable | Inconsistent
pub fn load(dir, label, prop)                -> io::Result<Option<Self>>   // never touches disk
pub fn load_and_quarantine(dir, label, prop) -> io::Result<Option<Self>>   // renames Undecodable bytes aside
```

Only `Undecodable` is quarantine-eligible — those bytes are the only surviving copy of the vectors. `Inconsistent` (decodes, fails `validate_invariants`) stays put, as before. `Unreadable` (dangling symlink) stays put too: there are no bytes to preserve, so renaming would destroy evidence and protect nothing.

| caller | uses |
|---|---|
| `helpers::load_vector_indexes` (open) | `load_and_quarantine` — it is about to take ownership of the slot, so its own later `save()` is what would overwrite the damaged bytes |
| `helpers::health` (both diagnostic tiers) | `load` |
| `expr.rs` `hybrid_vector_index` (query) | `load` |

**`sparrowdb`** — one report type, two tiers:

```rust
pub struct VectorIndexFailure { label, prop, path, reason }
pub struct VectorIndexHealth  { unscannable: Option<String>, active: Vec<_>, historical: Vec<_> }
impl VectorIndexHealth { pub fn is_healthy(&self) -> bool }   // unscannable.is_none() && active.is_empty()

GraphDb::vector_index_health(path)         // NEW. directory metadata only, no contents read
GraphDb::vector_index_load_failures(path)  // as before, plus a real load of every live file
```

**`sparrowdb-node` / npm** — the same three signals, unflattened:

```typescript
interface VectorIndexDamageReport {
  unscannable?: string                   // present when the directory could not be listed
  active: VectorIndexLoadFailure[]       // unrecovered damage — alert on this
  historical: VectorIndexLoadFailure[]   // recovered-from damage — forensics only
  healthy: boolean                       // !unscannable && active.length === 0, computed natively
}

SparrowDB.vectorIndexLoadFailures(path): VectorIndexDamageReport   // deep
SparrowDB.vectorIndexDamageScan(path):   VectorIndexDamageReport   // NEW, cheap — this is the monitor's call
```

`is_healthy()` is not a method across the napi boundary, so it is exposed as a precomputed `healthy` field, derived in Rust from `is_healthy()`. A JS caller cannot get the composition wrong, and the JS suite asserts `healthy === (!unscannable && active.length === 0)` on every report it receives.

The napi type is `VectorIndexDamageReport`, not `VectorIndexHealth`: that name is already taken by the **instance** method `db.vectorIndexHealth(label, prop)`, which counts reachable vectors *inside one healthy index*. Different question, different receiver; both doc comments now say so explicitly.

`unscannable` is `Option<String>`, which napi renders as an **absent** property rather than `null` — the `.d.ts` declares `unscannable?: string` and the docs say to test it for truthiness. This was verified by execution, not assumed.

### Why the cheap tier does not verify live files

`vector_index_health` / `vectorIndexDamageScan` cost one `read_dir` plus one `stat` per live entry — independent of index size. They report quarantine artifacts (split `active`/`historical`) and live entries that do not resolve, but do **not** check that a live index still decodes.

That is sound for a *running* store, and the reason is structural rather than convenient: `open` loads and validates every live index and refuses to start when one fails, so a store that is up has had every live file verified already. The residual question a monitor is asking is "did we quarantine something and lose vectors?", which is exactly what this answers.

I deliberately did **not** make the poll call a pure name scan. Dropping the deep live-arm would create a *new* instance of the blindness this PR is fixing: an index that decodes but fails `validate_invariants` is deliberately never quarantined, so a name-only diagnostic would report nothing for a store that refuses to open.

### Composition with #453

#453 landed first and adds two sidecars next to every index: `<index>.bin.lock`, never unlinked by design, and `<index>.bin.tmp.<pid>.<nonce>`, which a crashed writer leaves behind for good. Both sit in the directory this PR classifies by file name, so a wrong guess would break every store that has ever saved a vector.

The rebase was textually clean, including in `vector_index.rs`. Reading the two patches says they compose — but #442 and #446 also read as composing and produced #450, so it is asserted by execution in test 9 instead.

One direction worth recording: this change strictly **reduces** the number of racers against #453's new save lock. Before it, three code paths could rename an index file out from under a writer holding that lock — `open`, the diagnostic, and a `hybrid_search` read. Now only `open` can. `load_and_quarantine` still does not take the save lock, but that is unchanged from `main` and is neither introduced nor worsened here.

**Breaking changes:** `GraphDb::vector_index_load_failures` returned `Vec<(String, String, PathBuf, String)>`; `SparrowDB.vectorIndexLoadFailures` returned `VectorIndexLoadFailure[]`. Both now return the report object. Every in-tree consumer is migrated in this PR (`db.rs`, `crates/sparrowdb-node/src/lib.rs`, `npm/sparrowdb/index.d.ts`, `npm/sparrowdb/test/integration.test.js`, `crates/sparrowdb/tests/vector_index_load_errors.rs`).

## expr_decision — refuse, do not degrade

Three candidates. **Returning an error is not reachable**: `eval_expr_graph` and every scalar function it dispatches are infallible by signature (`-> Value`), and making them fallible is an engine-wide change through `engine/{mod,scan,path,hop}.rs` — far outside this fix. A separate issue should track it; the same swallow class is alive in FTS (`mutation.rs:944-948`, `db.rs:978-982`, `engine/mod.rs:274-282`, and `hybrid_search`'s own FTS branch) and would be fixed by the same plumbing.

That leaves degrade-to-FTS or refuse. **Refuse**, because degrading produces the *same observable* as a legitimately absent index — precisely the absent-vs-damaged confusion #445 exists to eliminate. A caller who asked for a fusion of vector and text, and silently got text only, has been told the store is healthy by a store that is not.

So the loader returns three states and the caller keeps them apart:

- **absent** (`Ok(None)`) → full-text-only, unchanged. A pair with no index is a legitimate configuration.
- **damaged** (`Err`) → `Value::Null` plus `tracing::warn!` naming label, property and decode reason.

`Value::Null` is already this function's failure signal (bad arity, bad argument types, fusion failure), so it introduces no new shape, and a caller cannot mistake it for an empty result list.

Two related fixes in the same loader: the fingerprint is now a cache-validity token rather than an existence test (it opens the path, so a dangling symlink reported `NotFound` and a damaged header reported an error — neither means "no index here"), and an index with no fingerprint is no longer cached, because there would be no token to invalidate it against.

### Both verdicts are cached, under the same token

`hybrid_search` is a scalar function: it is evaluated once per candidate row. The success path has been cached against the on-disk fingerprint since #442, but the failure path was not — so a damaged index was re-read and re-decoded on **every** row, and emitted an identical warning on every row.

For an 8 MB `Knowledge.embedding` index an N-row scan is N × 8 MB of I/O plus N identical log lines: an I/O storm and a log flood arriving exactly when the store is already degraded and someone is reading the logs to find out why. A file that decodes cleanly and then fails `validate_invariants` paid a full `bincode` decode every row. A change whose purpose is "behave well when the index is damaged" must not amplify the incident.

The damaged verdict is now cached as `CachedIndex::Damaged(reason)` under the same fingerprint token as the success, in the same bounded map, so invalidation is identical for both: rebuild the index, the fingerprint changes, the entry stops matching, the next row sees the repair. `Value::Null` is unchanged as the observable — the absent-vs-damaged distinction is the point.

**A verdict is cached only when a fingerprint exists**, because the fingerprint *is* the invalidation token. That costs nothing: `fingerprint` fails only when the file cannot be opened at all (a dangling symlink, a permission block), and those re-probe for one failed `open` plus one `lstat` with no payload read — never the full decode this cache exists to avoid. Every case that costs a decode has a fingerprint. Caching a negative without a token could let it outlive its repair, recreating the "fixed but still reports broken" failure this PR removes from the diagnostic, one layer down; a repeated log line is the better trade in the one case where the two conflict.

The bound is now stated rather than implicit: the cache is process-global and keyed by `(dir, label, prop)`, so a long-running process — an MCP server that opens many databases is the realistic case — would otherwise retain one entry per pair it had ever queried, each `Ready` entry pinning a fully deserialised multi-megabyte index. Capped at 8, cleared wholesale: there is no access-recency information worth tracking, and a cleared cache costs a reload, not a wrong answer.

## issues_closed

**#456 — yes.** The query path no longer mutates and no longer swallows; the diagnostic no longer mutates; all sub-findings addressed.

**#455 — yes.** #455 is that the live arm reports the `.bin` path while `load` renames that file during the same probe, so `fs.existsSync(entry.path)` is `false` on the first — and, given #451, most-acted-on — report. Measured at the JS boundary against the pre-fix tree: `fs.existsSync(entry.path)` → **false**, one artifact created by a *read*, and the second call returned something different from the first. Post-fix all three are fixed, and `npm/sparrowdb/test/integration.test.js` asserts each. #455 suggested reporting the quarantine destination instead; removing the rename achieves the stated goal and also fixes the second-call-differs half.

**#459 — filed, not fixed here, and confirmed PRE-EXISTING.** Found while building the negative-cache test: `MATCH (d:Doc) RETURN hybrid_search(...)` returns `Null` for a **healthy** index, because the chunked pipeline never routes `Expr::FnCall` through the graph-aware evaluator. Same class as #415/#416.

Verified by execution against unmodified c9f7a0b from an isolated target dir, not inferred from the diff: all six query shapes behave **identically** on main and on this branch, row for row. It is therefore pre-existing and ships in 0.1.22 — not introduced here. The full comparison table is on #459. The fix belongs in `engine/{scan,pipeline_exec}.rs`, which this PR does not touch.

**#451 — not closed.** Fail-closed `open()` is still one-shot: after quarantine the store reopens "healthy" with the pair silently unconfigured. This PR reduces the restart count from N+1 to 2 and keeps the artifacts visible in `active`, but that behaviour is unchanged and out of scope.

## sub_findings_fixed

**1. N damaged indexes required N failed restarts** — fixed. `load_vector_indexes` `return`ed on the first `Err`; it now runs to completion and reports every failure in one `Error::Corruption`. With two planted, **pre-fix the message named only `(Doc, embedding)`** — the scan is sorted by `(label, prop)` and `"Doc" < "Note"`, so `Note` was never probed — and one artifact was created. Post-fix the message names both, both are quarantined by the same open, and the second open succeeds: 2 restarts, not 3.

**2. Stale artifacts false-flagged a healthy store forever** — fixed with `historical`, not suppression. An artifact whose `(label, prop)` is served by a working index moves to `historical`; one with nothing serving its pair stays in `active`. This is the adoption-deciding property: artifacts are never cleaned up automatically, so a report that keeps naming a repaired pair can only ever go green → red with no path back, and an alert that never clears after the problem is fixed gets muted — worse than no alert, because the attention has been spent and nothing was bought with it.

Reported separately rather than suppressed because #442 records no reason at quarantine time and writes no sidecar: those bytes are the only surviving evidence that the incident happened.

This does not weaken #450 — the artifact-is-the-only-file case is still `active`. Only the case where a working index sits beside it changed.

Stated plainly: neither tier can tell whether a rebuilt index holds everything the artifact held. Nothing can — the artifact does not decode, which is why it is an artifact. Staying red forever would not have asserted that either.

**3. Dangling symlink classified as absent** — fixed. `load` began with `if !path.exists()`, which follows symlinks and is `false` for a broken one → `Ok(None)` → "no index configured". `read_index` now distinguishes with `symlink_metadata`; the diagnostic makes the same distinction with one `stat` at both tiers. **Pre-fix: 0 entries reported and `open()` → `Ok`** with the pair silently unconfigured. Post-fix: 1 active failure at both tiers, `open()` refuses.

**4. `[]` conflated "healthy" with "could not tell"** — fixed. `scan_vector_index_dir` now returns `Result<Vec<IndexFile>, String>`; an absent directory is still `Ok(vec![])` (distinguished with `symlink_metadata`, so a dangling symlink standing in for the directory is *not* absent), while a permission block, a plain file in the directory's place, an I/O error, or an unreadable entry becomes `unscannable`. `is_healthy()` is false for it: not knowing is not the same as being fine. Still no `Err` channel — a diagnostic must not become a failure path.

`open()` also now refuses on an unlistable `vector_indexes/`, for the same fail-closed reason it refuses on a damaged file: a directory we cannot list may hold indexes, and starting anyway means "no index configured" for every one of them.

**Deferred deliberately:** #451 (above); the FTS swallow class (needs the fallible-evaluator plumbing); automatic cleanup of `historical` artifacts (deleting the only evidence of an incident is the operator's call).

### Rust — `crates/sparrowdb/tests/regression_456_load_is_not_destructive.rs`, 10 tests

Every expected value hand-derived in a comment from the fixture the test builds; none captured from output.

1. **`hybrid_search_on_an_index_damaged_after_open_neither_renames_nor_reports_zero_hits`** (headline). One `Doc` node, one 3-d cosine vector, no FTS index. Baseline derivation: a 1-element index searched with `k*2 = 2` returns exactly that node id; `FtsIndex::open` fails with no FTS index so the text list is empty; RRF over (1-element, empty) yields one entry and `truncate(k=1)` leaves it — exactly `[node_id]`. Then truncate to 4 bytes with the db still open. The cache miss is derived *exactly*, not probabilistically: the healthy v2 fingerprint is `(generation, crc32c as u64)` whose second element is ≤ `u32::MAX` = 4_294_967_295, while the 4-byte file falls back to `(file_len, mtime_nanos)` ≈ `(4, 1.7e18)` — greater than `u32::MAX` for every instant after 1970-01-01 00:00:04.295, so the tuples can never compare equal. Asserts the file is present, byte-identical, no artifact created, and the value is `Null` and specifically not a `List`.
2. **`one_open_reports_every_damaged_index`**. Two indexes truncated to 4 bytes. Two files, two distinct pairs, both undecodable → one open names both, quarantines both, next open succeeds with neither pair configured; then 2 artifacts with nothing serving either pair → 2 `active`, 0 `historical`, sorted `[("Doc","embedding"), ("Note","vec")]`.
3. **`a_rebuilt_pair_moves_from_active_to_historical`**. Healthy `(Memory, embedding)` + an artifact for that same pair + an artifact for `(Ghost, vec)` with no live file. One pair is served → its artifact is debris → 1 `historical`; the other has nothing serving it → 1 `active`. Asserted at **both** tiers, since neither conclusion needs a file's contents. Removing the orphan then makes the store `is_healthy()` with `historical` still at 1.
4. **`the_diagnostic_is_read_only_and_reports_a_path_that_resolves`**. One truncated live file, never opened → 1 `active` from the live arm whose path is that `.bin`. Asserts `path.is_file()`, no artifact created, bytes unchanged, `second == first`.
5. **`load_leaves_the_file_alone_and_load_and_quarantine_moves_it`**. Both entry points return `Err` on 4 bytes; `load` leaves the file (twice, to show it is repeatable); `load_and_quarantine` moves it, and the artifact is byte-identical to the damaged bytes.
6. **`a_dangling_symlink_is_not_mistaken_for_an_absent_index`** (unix). `symlink_metadata` succeeds, `exists()` false → 1 `active` at both tiers, `open()` refuses, nothing quarantined.
7. **`an_unlistable_directory_is_not_reported_as_healthy`** (unix). Healthy control captured first, then `chmod 000` → `unscannable.is_some()`, both lists empty, `!is_healthy()`, `health != control`, and `open()` refuses. Skips with a printed note if `chmod 000` does not block the uid (root), so it never asserts something untrue of the run.
8. **`the_cheap_tier_reads_no_index_contents`** (unix). `chmod 000` on the index *file* — `stat` still succeeds, `open` does not. Cheap tier stats and stops → healthy; deep tier gets `EACCES` → exactly 1 `active` naming that pair and path. The two tiers can only disagree if the cheap one never opened the file.
9. **`save_sidecars_from_issue_453_are_not_mistaken_for_indexes`** (new, from the rebase). #453 puts `<index>.bin.lock` — which it deliberately never unlinks — and `<index>.bin.tmp.<pid>.<nonce>` staging files next to every index. Neither is an index, and both live in the directory this PR classifies **by file name**: a scanner that guessed wrong would either refuse to open a healthy store or report one as damaged, on every store that has ever saved a vector. Derivation: the scanner recognises exactly two shapes, a name ending `.bin` and a name containing `.bin.corrupt.`; `…​.bin.lock` ends `.lock`, `…​.bin.tmp.99999.0` ends `.0`, `…​.bin.tmp` ends `.tmp`, and none contains `.corrupt.` — so all three are ignored and the only recognised entry is the healthy live index. Expected at both tiers: 0 active, 0 historical, healthy; and `open()` still returns the index with its one vector. The lock file is produced by a **real** `save()` rather than written by the test, and the test asserts it exists, so it fails loudly if #453 ever stops creating it instead of quietly testing nothing.
10. **`a_damaged_index_is_read_and_warned_about_once_per_incident_not_once_per_row`** (new). Five `Doc` nodes, one saved index, then damaged under the open database. The **warning count is the observable**: row 2 can only skip its warning by taking the cached path, so `warns == 1` over a five-row scan is direct evidence that four rows never touched the file. A `tracing` layer counts `WARN` events mentioning `hybrid_search`; it is installed globally rather than thread-locally because the engine may evaluate rows on a worker thread and a thread-local default would silently count zero — a test that cannot observe what it asserts is worse than no test. Then the original image is written back mid-session and every row must see the repair with zero warnings. That the cached negative cannot survive the repair is hand-derived, not assumed: the restored fingerprint is `(generation, crc32c)` whose second element is at most `u32::MAX` = 4_294_967_295, while the damaged file's fallback is `(4, mtime_nanos)` at ~1.7e18, so the two can never compare equal.

    The `WITH d` in the test query is **load-bearing**. Without it the chunked pipeline serves the scan and never dispatches `hybrid_search` at all, returning `Null` for a perfectly healthy index — the test would have passed for entirely the wrong reason, never reaching the path it claims to exercise. The baseline assertion is what caught that, and it is filed separately as #459.

### JavaScript — `npm/sparrowdb/test/integration.test.js`, 7 tests (4 rewritten, 3 new)

Run by the existing CI Node.js job. Every report received is checked with `assertReportShape`, which re-derives `healthy` from `unscannable` and `active.length` and asserts the native field agrees.

- **`both tiers are static methods on the class, not instance methods`** — extends the #454 guard to the new static, and pins that the *instance* `vectorIndexHealth(label, prop)` stays an instance method, so the two "health" calls cannot be confused.
- **`never throws, and distinguishes "nothing wrong" from "could not look"`** (new). Derived from `scan_vector_index_dir`: an absent `vector_indexes/` gives `read_dir` ENOENT *and* `symlink_metadata` ENOENT → ordinary no-index case → scanned, healthy. A path containing a NUL byte cannot reach the OS at all; `read_dir` fails `InvalidInput`, which is not `NotFound` → `unscannable`, `healthy === false`. Both tiers.
- **`a chmod-000 index directory is not reported as healthy`** (new). Healthy control, then `chmod 000` on `vector_indexes/`: `unscannable` is a string, `active` empty, `healthy === false`, `notDeepEqual(blind, control)`, and `open()` throws. Skips under root.
- **`a healthy database reports healthy, with both lists empty`** — control, both tiers.
- **`reports live damage WITHOUT quarantining it — the report is read-only`** (rewritten; this is the #455 guard). Asserts `active.length === 1`, `path === indexFile`, `fs.existsSync(entry.path)`, bytes unchanged, **zero** artifacts created by the read, and `second` deep-equals the first.
- **`a rebuilt pair moves from active to historical`** (new). Damaged index → failed `open()` quarantines it → 1 `active`, 0 `historical`. Operator calls `createVectorIndex` for the same pair → the artifact count is still exactly 1 (rebuilding must not touch it) and the *same* artifact must now be 1 `historical`, 0 `active`, `healthy === true`, with `fs.existsSync(historical[0].path)` still true. Both tiers.
- **`is callable on a database that refuses to open, and keeps reporting after it starts opening "clean"`** — the #451 guard from #454, migrated to `report.active`, with `healthy === false` added.

**Existing tests updated, not weakened:** `vector_index_durability.rs::rejected_index_file_is_quarantined_rather_than_left_to_be_overwritten` now calls `load_and_quarantine` (assertions unchanged — that arm still quarantines). `vector_index_load_errors.rs` migrated from tuple indexing to named fields; all six assertions kept.

## prefix_failure_proof

Proved by reverting the source to the pre-fix tree (`git stash` / `git checkout c9524be -- <files>`), running, then restoring. The Rust test file was restored byte-exact and verified by sha256 (`88ddfaf0…` round 1, `60a50ed6…` round 2).

**Round 1 — the six original Rust tests: 6 failed, 0 passed**, each with the named symptom. The only adaptation was `load_and_quarantine(` → `load(`, since the function does not exist pre-fix; that substitution makes test 5 fail on the right assertion.

```
hybrid_search_..._neither_renames_nor_reports_zero_hits  a READ QUERY renamed .../hnsw_Doc_embedding.bin away
one_open_reports_every_damaged_index                     `Note` missing from: corruption: vector index for (Doc, embedding) …
the_diagnostic_is_read_only_and_reports_a_path...        the reported path .../hnsw_Memory_embedding.bin does not resolve
a_stale_artifact_beside_a_healthy_index_is_not_reported  left: 2  right: 1
a_dangling_symlink_is_not_mistaken_for_an_absent_index   left: 0  right: 1
load_leaves_the_file_alone_and_load_and_quarantine...    load() must not move the file
```

**Round 2 — MEASURED pre-fix values for the reshaped and newly added Rust claims:**

```
P7 control (healthy store)             : len=0
P7 chmod-000 dir really blocked        : true
P7 blind (chmod 000 vector_indexes/)   : len=0
P7 blind == control                    : true          <- green and blind are byte-identical
P7 open() on unlistable dir            : Ok
P7   get_vector_index(Memory,embedding): false         <- and the index silently vanished
P8 chmod-000 file really blocked       : true
P8 pre-fix deep report                 : len=1
P8 pre-fix cheap tier                  : does not exist (no vector_index_health)
P3 rebuilt pair + orphan artifact      : reported=2 (expected active=1, historical=1)
P3   reported: (Ghost, vec)
P3   reported: (Memory, embedding)                     <- repaired pair still red
P6 dangling symlink reported           : len=0
P6 open() with dangling symlink        : Ok            <- present, treated as absent
```

**Round 3 — MEASURED at the JS boundary**, against a `--release` `sparrowdb-node` built from the pre-fix source:

```
J5 return type              : Array
J5 reported len             : 1
J5 reported path            : .bin
J5 fs.existsSync(entry.path): false     <- #455, at the boundary the consumer actually calls
J5 artifacts created by READ: 1         <- a getter-shaped call renamed the file
J5 second call == first     : false     <- and the first call caused the difference
J6 after rebuild, reported  : 1 (expected active=0, historical=1)
J6 has `historical` field   : no — bare array
J6 has `healthy` field      : no — bare array
J3 chmod really blocked     : true
J3 control                  : []
J3 blind                    : []
J3 blind === control        : true      <- green and blind are byte-identical in JS too
J3 has `unscannable` field  : no — bare array
J3 open() on unlistable dir : Ok
```

**Round 4 — MEASURED for the negative cache**, with only the caching change reverted (the rest of the PR in place), five-row scan against one damaged index:

```
warnings emitted, before the caching commit : 5     <- one per row, one full re-read per row
warnings emitted, after                     : 1
after mid-session repair                    : 0, and every row returns a list
```

Test 8 (`the_cheap_tier_reads_no_index_contents`) and the `vectorIndexDamageScan` assertions cannot "fail pre-fix" in the regression sense: the cheap tier did not exist. P8 and the J-rows record what the only pre-fix behaviour was.

## verification_run

`CARGO_TARGET_DIR=/Volumes/Dev/target-456`, cargo `/Users/ryaker/.rustup/toolchains/stable-aarch64-apple-darwin/bin/cargo`. All re-run on the tree rebased onto c9f7a0b.

| command | result |
|---|---|
| `cargo test -p sparrowdb -p sparrowdb-storage -p sparrowdb-execution --no-fail-fast` | exit 0 — see below |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --workspace --exclude sparrowdb-ruby --all-targets --keep-going -- -D warnings` | clean |
| `cargo build --release -p sparrowdb-node` (CI's Node.js job, reproduced locally) | builds |
| `node --test npm/sparrowdb/test/integration.test.js` | **47 pass, 0 fail** |
| fresh clone of the pushed branch, `cargo fmt --all -- --check` | clean |

`tracing-subscriber` is added as a **dev-dependency** of `sparrowdb` for the warning-count test. It was already a workspace dependency and already in `Cargo.lock`, so nothing new is vendored and no shipped artifact gains a dependency.

## could_not_verify

- **Anything requiring `cargo test --workspace`.** `sparrowdb-ruby` does not compile against the installed Ruby here (magnus/rb_sys, ~79 errors). It was not touched and has no reference to these APIs, but it was not compiled.
- **`index.d.ts` does not parse as TypeScript** — pre-existing, issue #449. I matched the surrounding style rather than fixing it, so my additions inherit the same problem. Not verified with `tsc`.
- **Non-unix behaviour.** Rust tests 6-8 and the JS `chmod` test are unix-gated or unix-dependent. Dangling-symlink and permission semantics on Windows are unverified; the underlying `symlink_metadata` / `read_dir` calls are portable but the fixtures are not. The CI Node.js job runs on Linux only, so the `chmod` test is not exercised on Windows.
- **Running as root.** The three permission tests skip themselves if `chmod 000` does not block the uid. On a root runner they pass without exercising anything.
- **The 8 MB scale claim.** Argued from the code path (`load` → `fs::read` + `bincode::deserialize` + `validate_invariants` over the whole graph, once per live index per call) and from the consumer's report. Not benchmarked on an 8 MB index.
- **Concurrency.** No test exercises a diagnostic running concurrently with a writer's `save()`, or two `open()`s racing to quarantine the same file. The scan-then-probe window is unchanged from `main`.
- **Concurrency against #453's save lock.** Test 9 covers the static composition — that the new `.lock` and `.tmp.<pid>.<nonce>` sidecars are not misread as indexes. It does **not** cover a diagnostic running while a writer holds the save lock, nor two `open()`s racing to quarantine the same file. `load_and_quarantine` does not take that lock; unchanged from `main`, but unverified under contention.
- **Which thread evaluates a row.** The warning counter uses a global `tracing` subscriber precisely so the count is correct either way, but I did not confirm whether rows are evaluated on the calling thread or a worker. The assertion holds regardless; the claim I am *not* making is that evaluation is single-threaded.
- **npm publish artifacts.** The `@sparrowdb/*` platform packages are not rebuilt here; only the locally built `.node` was exercised.

### DIFF
--- a/crates/sparrowdb-execution/src/engine/expr.rs
+++ b/crates/sparrowdb-execution/src/engine/expr.rs
@@ -162,13 +162,26 @@ impl Engine {
     /// same write lock that mutates it, a matching fingerprint means the cached
     /// copy is byte-identical to what the current writer would hand us.
     ///
-    /// Returns `None` when the index does not exist or fails to load; the
-    /// caller then falls back to full-text-only results, exactly as before.
+    /// # Absent is not damaged (#456)
+    ///
+    /// `Ok(None)` means no index is configured for the pair — a legitimate
+    /// state, and the caller degrades to full-text-only results exactly as
+    /// before.  `Err(reason)` means a file *is* there and cannot be served.
+    ///
+    /// Collapsing those two into `None` is what made this a data-loss path: the
+    /// loader used to be `VectorIndex::load(..).ok().flatten()?`, and since #442
+    /// had moved the quarantine rename inside `load`, a plain read query
+    /// renamed the last live copy of the index aside, then reported success
+    /// with zero vector hits.  This function now calls the non-destructive
+    /// `load` and keeps the two outcomes apart.
     fn hybrid_vector_index(
         vec_dir: &std::path::Path,
         label: &str,
         prop: &str,
-    ) -> Option<std::sync::Arc<sparrowdb_storage::vector_index::VectorIndex>> {
+    ) -> std::result::Result<
+        Option<std::sync::Arc<sparrowdb_storage::vector_index::VectorIndex>>,
+        String,
+    > {
         type Key = (std::path::PathBuf, String, String);
         type Entry = (
             (u64, u64),
@@ -177,35 +190,71 @@ impl Engine {
         static CACHE: std::sync::OnceLock<std::sync::Mutex<HashMap<Key, Entry>>> =
             std::sync::OnceLock::new();
 
-        // Missing file / unreadable header → nothing to search.
+        // The fingerprint is a cache-validity token, not an existence test.
+        // `fingerprint` opens the path, so a dangling symlink reports NotFound
+        // and a damaged header reports an error — neither of which means "no
+        // index here".  When it yields nothing we fall through to `load`, which
+        // is the one place that distinguishes absent from unreadable.  That
+        // costs two extra syscalls per row in the no-index-configured case,
+        // against the full `FtsIndex::open` the same row already performs.
         let fingerprint =
             sparrowdb_storage::vector_index::VectorIndex::fingerprint(vec_dir, label, prop)
-                .ok()
-                .flatten()?;
+                .unwrap_or_default();
 
-        let cache = CACHE.get_or_init(|| std::sync::Mutex::new(HashMap::new()));
-        let mut guard = cache.lock().ok()?;
         let key: Key = (vec_dir.to_path_buf(), label.to_owned(), prop.to_owned());
+        let cache = CACHE.get_or_init(|| std::sync::Mutex::new(HashMap::new()));
+        let mut guard = match cache.lock() {
+            Ok(g) => g,
+            // A poisoned cache mutex is not evidence about the index file.
+            // Skip the cache rather than claim the index is missing.
+            Err(_) => {
+                return match sparrowdb_storage::vector_index::VectorIndex::load(
+                    vec_dir, label, prop,
+                ) {
+                    Ok(idx) => Ok(idx.map(std::sync::Arc::new)),
+                    Err(e) => Err(e.to_string()),
+                }
+            }
+        };
 
-        if let Some((cached_fp, idx)) = guard.get(&key) {
-            if *cached_fp == fingerprint {
-                return Some(std::sync::Arc::clone(idx));
+        if let Some(fp) = fingerprint {
+            if let Some((cached_fp, idx)) = guard.get(&key) {
+                if *cached_fp == fp {
+                    return Ok(Some(std::sync::Arc::clone(idx)));
+                }
             }
         }
 
-        let loaded = sparrowdb_storage::vector_index::VectorIndex::load(vec_dir, label, prop)
-            .ok()
-            .flatten()?;
+        let loaded = match sparrowdb_storage::vector_index::VectorIndex::load(vec_dir, label, prop)
+        {
+            Ok(Some(idx)) => idx,
+            // The pair no longer has a usable index.  Drop any cached copy
+            // rather than keeping an 8 MB deserialised graph alive for a key
+            // that can no longer be served.
+            Ok(None) => {
+                guard.remove(&key);
+                return Ok(None);
+            }
+            Err(e) => {
+                guard.remove(&key);
+                return Err(e.to_string());
+            }
+        };
         let idx = std::sync::Arc::new(loaded);
 
-        // Bound the cache: a process that opens many databases should not
-        // retain every index it has ever queried.
-        const MAX_CACHED_INDEXES: usize = 8;
-        if guard.len() >= MAX_CACHED_INDEXES && !guard.contains_key(&key) {
-            guard.clear();
+        // Only a fingerprinted index is cacheable: without one there is no
+        // token to invalidate the entry against, and a stale copy would be
+        // served forever.
+        if let Some(fp) = fingerprint {
+            // Bound the cache: a process that opens many databases should not
+            // retain every index it has ever queried.
+            const MAX_CACHED_INDEXES: usize = 8;
+            if guard.len() >= MAX_CACHED_INDEXES && !guard.contains_key(&key) {
+                guard.clear();
+            }
+            guard.insert(key, (fp, std::sync::Arc::clone(&idx)));
         }
-        guard.insert(key, (fingerprint, std::sync::Arc::clone(&idx)));
-        Some(idx)
+        Ok(Some(idx))
     }
 
     /// Evaluate `hybrid_search(label, emb_prop, text_prop, query_vec, query_text, k[, alpha])`.
@@ -273,12 +322,43 @@ impl Engine {
         };
 
         // ── 1. Vector search ────────────────────────────────────────────────
+        //
+        // A damaged index does NOT degrade to full-text-only here (#456).
+        //
+        // The three candidate behaviours were: return an error, quietly fall
+        // back to FTS, or refuse to answer.  Returning an error is not
+        // reachable from a scalar function — `eval_expr_graph` and every
+        // function it dispatches are infallible by signature, and making them
+        // fallible is an engine-wide change well outside this fix.  That leaves
+        // fall-back or refusal, and fall-back is the wrong one: it produces the
+        // *same* observable as a legitimately absent index, which is precisely
+        // the absent-vs-damaged confusion #445 exists to eliminate.  A caller
+        // who asked for a fusion of vector and text results, and silently got
+        // text only, has been told the store is healthy by a store that is not.
+        //
+        // So: absent → FTS-only, unchanged.  Damaged → `Value::Null`, which is
+        // already this function's failure signal (bad arity, bad argument
+        // types, fusion failure), plus a warning naming the file and the
+        // decode reason.  Null is a shape the caller cannot mistake for an
+        // empty result list.
         let vec_dir = self.snapshot.db_root.join("vector_indexes");
         let vec_results: Vec<(u64, f32)> =
             match Self::hybrid_vector_index(&vec_dir, &label, &emb_prop) {
                 // ef = max(k*2, 50) gives a reasonable exploration budget.
-                Some(idx) => idx.search(&query_vec, k * 2, (k * 2).max(50)),
-                None => vec![],
+                Ok(Some(idx)) => idx.search(&query_vec, k * 2, (k * 2).max(50)),
+                // No index configured for this pair: full-text-only is the
+                // honest answer and always has been.
+                Ok(None) => vec![],
+                Err(reason) => {
+                    tracing::warn!(
+                        label = %label,
+                        property = %emb_prop,
+                        reason = %reason,
+                        "hybrid_search: the vector index for this pair is present but unusable; \
+                         returning NULL rather than silently degrading to full-text-only results"
+                    );
+                    return Value::Null;
+                }
             };
 
         // ── 2. Full-text (BM25) search ───────────────────────────────────────
--- a/crates/sparrowdb-execution/src/engine/expr.rs
+++ b/crates/sparrowdb-execution/src/engine/expr.rs
@@ -1,6 +1,70 @@
 //! Auto-generated submodule — see engine/mod.rs for context.
 use super::*;
 
+/// Key of the process-global `hybrid_search` index cache: which database
+/// directory, which `(label, prop)` pair.
+type HybridCacheKey = (std::path::PathBuf, String, String);
+
+/// A cached verdict about one on-disk HNSW index, together with the fingerprint
+/// it was derived from.
+type HybridCacheEntry = ((u64, u64), CachedIndex);
+
+/// What the last look at an index file concluded — valid for exactly as long as
+/// that file's fingerprint is unchanged.
+enum CachedIndex {
+    /// The file loaded; this is the shared deserialised copy.
+    Ready(std::sync::Arc<sparrowdb_storage::vector_index::VectorIndex>),
+    /// The file is present and cannot be served, for this reason.
+    ///
+    /// # Why the negative is cached too (#456)
+    ///
+    /// `hybrid_search` is a scalar function: it is evaluated once per candidate
+    /// row.  Caching only the success meant a damaged index was re-read and
+    /// re-decoded on *every* row — for KMSmcp's 8 MB `Knowledge.embedding`
+    /// index that is N x 8 MB of I/O for an N-row scan, and a file that decodes
+    /// cleanly but then fails `validate_invariants` paid a full `bincode`
+    /// decode each time.  It also emitted one identical warning per row.
+    ///
+    /// Both arrive at the moment the store is already degraded and someone is
+    /// reading the logs to find out why, which is the worst possible time to
+    /// add an I/O storm and bury the signal.  A fix whose purpose is "behave
+    /// well when the index is damaged" must not amplify the incident.
+    Damaged(String),
+}
+
+/// A damaged-index verdict handed back to the caller.
+struct HybridIndexDamage {
+    /// The loader's own error text.
+    reason: String,
+    /// True only on the evaluation that first observed this damage for the
+    /// current fingerprint.  Every later row reuses the cached verdict and sets
+    /// this to `false`, so one damaged index produces one log line rather than
+    /// one per row.
+    first_observation: bool,
+}
+
+/// Insert into the `hybrid_search` index cache, keeping it bounded.
+///
+/// The cache is process-global and keyed by `(dir, label, prop)`, so without a
+/// bound a long-running process — an MCP server that opens many databases over
+/// its lifetime is the realistic case — would retain one entry for every pair it
+/// had ever queried, and each `Ready` entry pins a fully deserialised index that
+/// can be megabytes.  Eight is chosen to comfortably cover the pairs one query
+/// touches while keeping the retained set small; the eviction is a wholesale
+/// clear rather than an LRU because there is no access-recency information worth
+/// tracking here and a cleared cache costs one reload, not a wrong answer.
+fn hybrid_cache_insert(
+    map: &mut HashMap<HybridCacheKey, HybridCacheEntry>,
+    key: HybridCacheKey,
+    entry: HybridCacheEntry,
+) {
+    const MAX_CACHED_INDEXES: usize = 8;
+    if map.len() >= MAX_CACHED_INDEXES && !map.contains_key(&key) {
+        map.clear();
+    }
+    map.insert(key, entry);
+}
+
 impl Engine {
     // ── FTS scalar functions ──────────────────────────────────────────────────
 
@@ -166,7 +230,7 @@ impl Engine {
     ///
     /// `Ok(None)` means no index is configured for the pair — a legitimate
     /// state, and the caller degrades to full-text-only results exactly as
-    /// before.  `Err(reason)` means a file *is* there and cannot be served.
+    /// before.  `Err(..)` means a file *is* there and cannot be served.
     ///
     /// Collapsing those two into `None` is what made this a data-loss path: the
     /// loader used to be `VectorIndex::load(..).ok().flatten()?`, and since #442
@@ -174,21 +238,37 @@ impl Engine {
     /// renamed the last live copy of the index aside, then reported success
     /// with zero vector hits.  This function now calls the non-destructive
     /// `load` and keeps the two outcomes apart.
+    ///
+    /// # Both verdicts are cached, under the same token
+    ///
+    /// A damaged file is remembered as [`CachedIndex::Damaged`] against the same
+    /// fingerprint the successful load is cached against, so an N-row scan reads
+    /// and decodes it once rather than N times, and warns once rather than N
+    /// times.
+    ///
+    /// The invalidation is therefore identical for both: rebuild the index and
+    /// its fingerprint changes, the cached entry stops matching, and the next
+    /// row re-probes and sees the repair.  A negative verdict that outlived its
+    /// repair would recreate, one layer down, exactly the "fixed but still
+    /// reports broken" failure this PR removes from the diagnostic.
+    ///
+    /// A verdict is only cached when a fingerprint exists, because the
+    /// fingerprint *is* the invalidation token and there is nothing safe to do
+    /// without one.  That costs nothing in practice: `fingerprint` fails only
+    /// when the file cannot be opened at all (a dangling symlink, a permission
+    /// block), and those re-probe for one failed `open` plus one `lstat` with no
+    /// payload read — never the full decode this cache exists to avoid.
     fn hybrid_vector_index(
         vec_dir: &std::path::Path,
         label: &str,
         prop: &str,
     ) -> std::result::Result<
         Option<std::sync::Arc<sparrowdb_storage::vector_index::VectorIndex>>,
-        String,
+        HybridIndexDamage,
     > {
-        type Key = (std::path::PathBuf, String, String);
-        type Entry = (
-            (u64, u64),
-            std::sync::Arc<sparrowdb_storage::vector_index::VectorIndex>,
-        );
-        static CACHE: std::sync::OnceLock<std::sync::Mutex<HashMap<Key, Entry>>> =
-            std::sync::OnceLock::new();
+        static CACHE: std::sync::OnceLock<
+            std::sync::Mutex<HashMap<HybridCacheKey, HybridCacheEntry>>,
+        > = std::sync::OnceLock::new();
 
         // The fingerprint is a cache-validity token, not an existence test.
         // `fingerprint` opens the path, so a dangling symlink reports NotFound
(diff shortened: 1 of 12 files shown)


ryaker:
@coderabbitai review

The check reports `pass` with the body "Review rate limited", so no review has actually run against this diff. Requesting one.

Worth flagging for the reviewer: this is a **breaking API change** in both Rust and JS (`vector_index_load_failures` returns `VectorIndexHealth` rather than a 4-tuple; napi mirrors it as `VectorIndexDamageReport`), and it lands on top of #453, which puts two new sidecar files next to every index (`.bin.lock`, `.bin.tmp.<pid>.<nonce>`) that this PR's name-based classifier must not mistake for damage.

PR #463, opened 2026-08-03
fix(storage): bound WAL props count before allocating (closes #461)

### LINKED ISSUE #461: CRITICAL: opening an encrypted WAL with the wrong key aborts the process — unvalidated u32 count drives a 126 GiB allocation (wal/codec.rs:188)
## A wrong decryption key crashes the process instead of returning an error

`spa_98_wal_encryption` aborts on Linux CI:

```
running 4 tests
memory allocation of 135243999024 bytes failed
stack backtrace:
memory allocation of 180977756016 bytes failed
skipping backtrace printing to avoid potential recursion
error: test failed, to rerun pass `-p sparrowdb --test spa_98_wal_encryption`
Caused by:
  process didn't exit successfully: ... (signal: 6, SIGABRT: process abort signal)
```

126 GiB and 168.5 GiB. `SIGABRT`, not an `Err`.

## Cause — `crates/sparrowdb-storage/src/wal/codec.rs:186-188`

```rust
let count = u32::from_le_bytes(bytes[pos..pos + 4].try_into().unwrap()) as usize;
pos += 4;
let mut props = Vec::with_capacity(count);   // <-- allocates before any validation
```

`count` is an **unvalidated `u32` read straight from the buffer**. Decrypting with the wrong key produces garbage, so `count` becomes an arbitrary value up to 4.29e9. `WalProp = (String, Vec<u8>)` is 48 bytes on 64-bit, so:

| observed | ÷ 48 | implied `count` |
|---|---|---|
| 135,243,999,024 | 2,817,583,313 | plausible garbage `u32` |
| 180,977,756,016 | 3,770,369,917 | plausible garbage `u32` |

Both land under `u32::MAX`, and both are exactly `count × 48`. The mechanism is confirmed arithmetically, not inferred.

## The irony: the loop body is careful, the allocation is not

Every single field read inside the loop is bounds-checked and returns `Error::Corruption` — `props count truncated`, `prop key_len truncated`, `prop key truncated`, `prop val_len truncated`, `prop val truncated`. The code is meticulous about *reads*.

But `Vec::with_capacity(count)` runs **before the loop**, so none of that care is ever reached. The allocation is the one place a length is trusted, and it is the one place that aborts.

## Impact

- **Wrong password → process abort.** For an embedded database, an operator mistyping a key should get a clean error, not `SIGABRT` taking down the host process. There is no `catch_unwind` for an allocation failure.
- **Any corrupt WAL does this too**, not just encrypted ones — the same decoder runs on plaintext WAL replay. A single flipped byte in that `u32` is enough.
- Reachable during `open()`, so it is on the recovery path — the path that runs precisely when something has already gone wrong.

## Fix

`count` cannot legitimately exceed the bytes remaining. Each prop needs at least 8 bytes on the wire (`key_len` + `val_len`), so:

```rust
let remaining = bytes.len() - pos;
if count > remaining / 8 {
    return Err(Error::Corruption("props count exceeds remaining bytes".into()));
}
let mut props = Vec::with_capacity(count);
```

Or simply `Vec::new()` and let it grow — the allocation is amortised and this is not a hot path.

**Audit the same pattern elsewhere in the same file** while fixing: `codec.rs:232`, `:376`, `:410`, `:421`, `:506` and `:546` all read a length or count from the buffer. `:506` (`Vec::with_capacity(total_len)`) is the next most suspicious.

## Why this was invisible

It is the failing check on **#412**, the PR that adds `pipefail` so test failures actually fail the build. Before #412, `cargo test ... | tee` exited with `tee`'s status and this abort was reported **green**. This bug has been crashing CI silently for as long as it has existed.

Fixing it is on the critical path to merging #412 and getting a real test gate.

## Related

- #412 — the CI gate whose red status exposed this
- [reference_ci_pipefail_masking] — the masking mechanism

ryaker:
## Summary

`decode_props()` in `crates/sparrowdb-storage/src/wal/codec.rs:186` read an unvalidated `u32` count straight from the WAL buffer and passed it to `Vec::with_capacity(count)` **before** any of the loop's own bounds checks ran. Decrypting an encrypted WAL with the wrong key (or replaying any corrupt WAL — encryption isn't required to trigger this) turns `count` into garbage up to `u32::MAX`. With `WalProp = (String, Vec<u8>)` at 48 bytes, that's up to a ~192 GiB allocation attempt, which SIGABRTs the process instead of returning the `Error::Corruption` every other field read in this decoder already produces.

## Mechanism — measured, not inferred

The GH issue observed `memory allocation of 135243999024 bytes failed` and `180977756016 bytes failed` on Linux CI. Both divide evenly by `size_of::<WalProp>() = 48`, landing on plausible garbage `u32` values.

I reproduced the exact abort locally via a Linux container with strict overcommit accounting (`vm.overcommit_memory=2`, the setting that makes an oversized `mmap` fail immediately instead of being lazily reserved):

```
$ ./repro   # Vec::<WalProp>::with_capacity(u32::MAX) under vm.overcommit_memory=2
memory allocation of 206158430160 bytes failed
Aborted
$ echo $?
134   # SIGABRT
```

`206158430160 / 48 = 4294967295 = u32::MAX` exactly — confirms the arithmetic mechanism byte-for-byte, not just by analogy to the issue's numbers.

Running the **fixed** logic (bound check applied) under the identical strict-overcommit condition:

```
$ ./repro_fixed
REJECTED before allocating: count=4294967295 exceeds remaining/8=0
$ echo $?
0
```

**macOS caveat (this dev machine):** `Vec::with_capacity` for the same size does *not* abort on macOS — its VM overcommit lazily reserves address space without touching physical pages, so the reservation silently succeeds and the loop's own downstream bounds check then catches the corruption anyway (with a different `Error::Corruption` message). `spa_98_wal_encryption` passes 4/4 on this machine both **before and after** this fix, for the same reason — confirmed, not assumed. `ulimit -v` does not work on Darwin (`setrlimit failed: invalid argument`), so I used a Linux container with `vm.overcommit_memory=2` to get an apples-to-apples repro of the actual CI failure mode instead.

## Fix

```rust
let remaining = bytes.len() - pos;
if count > remaining / 8 {
    return Err(Error::Corruption(format!(
        "props count {count} exceeds remaining bytes ({remaining})"
    )));
}
let mut props = Vec::with_capacity(count);
```

Each prop needs >= 8 bytes on the wire (`key_len` + `val_len`), so `count > remaining / 8` is provably corrupt and rejected up front — same `Error::Corruption` type and style as the neighbouring checks in this file, no behavior change beyond turning an abort into the error it should already have been.

## Audit of every other length/count read in codec.rs

| Site | Reads | Disposition |
|---|---|---|
| `:186` (`decode_props`, props count) | **Fixed** — unbounded, drove the allocation directly |
| `:232` (`decode_str`, string length) | Validated against `bytes.len()` before the slice/`to_vec()`; safe |
| `:376` (`Write` payload, `image_len`) | Validated (`bytes.len() < 12 + image_len`) before slicing; safe |
| `:410` / `:421` (`NodeUpdate`, `before_len`/`after_len`) | Both validated against remaining bytes before slicing; safe |
| `:506` (`WalRecord::encode`, `Vec::with_capacity(total_len)`) | Write/encode path — `total_len` derives from an in-memory `Vec` the caller already owns, not an untrusted on-disk buffer. Not a corruption vector. |
| `:546` / `:622` (`WalRecord::decode`/`decode_with_version`, `length_val`) | Validated against `bytes.len()` before use; only slices (borrows), never allocates from the value. Safe |

`:186` was the only unguarded site in this file.

## Tests

Three new tests in `codec.rs`:
- `test_decode_props_rejects_bogus_count_before_allocating` — `decode_props` given an 8-byte buffer (`count=u32::MAX` + 4 trailing bytes). Expected outcome hand-derived: `remaining=4`, `4/8=0`, so `u32::MAX > 0` is rejected before any allocation.
- `test_node_create_rejects_bogus_props_count_via_public_entry_point` — same corruption exercised through the real `WalPayload::decode_plaintext` entry point WAL replay calls.
- `test_node_create_round_trip_with_props` — sanity check the new bound doesn't reject legitimate encodings (real prop lists are always >= 8 bytes/entry on the wire).

All three complete in <0.01s with no gigabyte-scale allocation (~245MB peak RSS measured).

## Test plan

- [x] `cargo test -p sparrowdb-storage --lib wal::codec` — 15 passed, 1 ignored (fixture generator), 0 failed
- [x] `cargo test -p sparrowdb --test spa_98_wal_encryption` — 4/4 passed (both before and after fix, on macOS — see caveat above)
- [x] `cargo test --workspace --exclude sparrowdb-ruby --no-fail-fast` — 216/216 test binaries `ok`, 1464 test cases, 0 failures
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --workspace --exclude sparrowdb-ruby --all-targets --keep-going -- -D warnings` — clean
- [x] Literal pre-fix repro on Linux with `vm.overcommit_memory=2`: SIGABRT, exit 134, "memory allocation of 206158430160 bytes failed"
- [x] Literal post-fix repro under identical conditions: clean rejection, exit 0

Closes #461.

Claude-Session: https://claude.ai/code/session_01Aq3C5J646sBCH7sVs3uaCE

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when decoding stored data to detect corrupt or implausibly large property counts.
  * Prevented invalid data from triggering excessive memory allocation.
  * Added coverage for valid records and corruption handling across decoding paths.

### DIFF
--- a/crates/sparrowdb-storage/src/wal/codec.rs
+++ b/crates/sparrowdb-storage/src/wal/codec.rs
@@ -185,6 +185,17 @@ fn decode_props(bytes: &[u8], offset: usize) -> Result<(Vec<WalProp>, usize)> {
     }
     let count = u32::from_le_bytes(bytes[pos..pos + 4].try_into().unwrap()) as usize;
     pos += 4;
+    // `count` is read straight from the (possibly corrupt/mis-decrypted) buffer
+    // and must be bounded before it drives an allocation. Every prop needs at
+    // least 8 bytes on the wire (key_len + val_len), so a count implying more
+    // props than could possibly fit in the remaining bytes is provably corrupt
+    // — reject it here rather than trusting it to size a `Vec`.
+    let remaining = bytes.len() - pos;
+    if count > remaining / 8 {
+        return Err(Error::Corruption(format!(
+            "props count {count} exceeds remaining bytes ({remaining})"
+        )));
+    }
     let mut props = Vec::with_capacity(count);
     for _ in 0..count {
         if bytes.len() < pos + 4 {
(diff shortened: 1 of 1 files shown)


ryaker:
@coderabbitai review

Focus areas for this one:

1. The bound is `count > remaining / 8`, justified by each `WalProp` needing at least 8 bytes on the wire (`key_len` + `val_len`). Please check that lower bound is actually correct for every encoding path — if any prop can be encoded in fewer than 8 bytes, this rejects legitimate data.
2. `decode_props` is on the `open()` recovery path, so a false rejection turns a recoverable database into an unopenable one. The failure mode of a too-strict bound is worse than the one of a too-loose one.
3. The audit concluded `:186` was the only site where a buffer-derived value reaches an allocation before validation. A second opinion on `:506` (`Vec::with_capacity(total_len)` in `WalRecord::encode`) and `:546`/`:622` would be welcome.

PR #465, opened 2026-08-03
fix(vector): quarantine() re-validates under SaveLock before renaming (closes #464)

### LINKED ISSUE #464: CRITICAL: quarantine() renames without the save lock or a re-check — a successful repair is silently destroyed, and the 'corrupt' artifact is the good file
## The race

`load_and_quarantine()` (`vector_index.rs:621-647`) reads the file, decides, and *then* renames — with nothing held across the gap:

```rust
match Self::read_index(&path)? {                  // (1) reads bytes, judges them
    LoadOutcome::Undecodable(reason) => {
        let quarantined = Self::quarantine(&path); // (2) renames whatever is at `path` NOW
```

And `quarantine()` (`:843-855`) is a bare rename — **no lock, no re-read, no identity check**:

```rust
fn quarantine(path: &Path) -> Option<PathBuf> {
    let dest = /* <path>.corrupt.<millis> */;
    match std::fs::rename(path, &dest) { ... }
}
```

Interleaving:

| | process A (`open()`) | process B (repair) |
|---|---|---|
| T0 | | index at `path` is corrupt |
| T1 | `read_index` reads the corrupt bytes → `Undecodable` | |
| T2 | | `save()` takes `SaveLock`, writes `<path>.tmp.<pid>.<nonce>`, **`rename` → `path`**. Repair succeeds. |
| T3 | `quarantine(&path)` renames **the new, good file** to `<path>.corrupt.<stamp>` | |

**The repair is silently undone.** Afterwards the pair has no live index — it reads as *absent* — and the file sitting in quarantine is a perfectly valid index.

## Why this is worse than an ordinary lost update

The quarantine artifact is the **evidence trail**. An operator investigating an incident opens `<path>.corrupt.<stamp>`, finds a completely healthy index, and concludes the tooling is lying. #442 records no reason at quarantine time, so there is nothing in the artifact to say which file it actually was.

It also defeats the remediation loop shipped in #457: repair → healthy → concurrent `open()` → destroyed → `active` again. The monitor does at least go red rather than silently green, so this is visible, but the repair itself does not survive.

## The lock already exists and quarantine simply does not take it

#453 added `SaveLock` — a real `flock(2)` advisory lock on `<path>.lock`, documented at `:111-125` and taken by `save()` at `:412`. It is exactly the right primitive. `quarantine()` predates it and never acquires it.

## Fix

`quarantine()` must (a) take the same `SaveLock`, and (b) **re-validate under the lock before renaming** — because the bytes it renames must be the bytes it judged. Acquiring the lock alone is not sufficient: A could still block on the lock, acquire it after B's save completes, and then rename the good file. The re-check is what closes the window.

Cheapest correct form: capture the file's identity at read time — the `(generation, crc32c)` fingerprint already used for the negative cache — and rename only if it still matches under the lock. If it does not match, someone replaced the file; return `Absent`/retry the load rather than quarantining.

## Test

Two processes, deterministic ordering (the `regression_452.rs` multi-process harness is the pattern):
- A pauses between `read_index` and `quarantine`
- B completes a full repair `save()`
- A resumes

Assert the good index is **still at `path`**, no `.corrupt.` artifact was created, and `vector_index_load_failures()` reports `healthy`. Derive expected values by hand; confirm the test fails against `5f36b02`.

## Provenance

`#458` correctly removed the destructive read path — `load()` is now non-destructive and only `open()` quarantines. That closed the large hole. This is the remainder its author explicitly flagged and did not claim to fix: *"`load_and_quarantine` still does not take the save lock, but that is unchanged from main and is not something this PR introduces or worsens."* Correct on both counts, and it is still live.

## Related

- #453 — introduced `SaveLock`; `quarantine()` does not take it
- #458 — removed read-path quarantine; left this
- #457 — the remediation loop this defeats
- #451 — still open; a store reopens "healthy" after a quarantine

ryaker:
## Summary

`load_and_quarantine()` judged a file with one unlocked `read_index`, then `quarantine()` renamed whatever was at `path` with **no lock and no re-check**. Between the judgement and the rename, a concurrent `save()` could complete a repair. The rename then fired on whatever was on disk *at that moment* — the newly repaired, healthy index — instead of the corrupt bytes it was supposed to preserve. Net effect: the repair was silently undone, the pair read as absent, and the file sitting in `*.corrupt.*` — the only place an operator would look for an explanation — was a perfectly healthy index.

## Fix

`quarantine_after_recheck()` (new) sits between `load_and_quarantine`'s unlocked judgement and the rename:

1. **Takes `SaveLock`** — the same `flock(2)` advisory lock `save()` already takes for its whole read-modify-write (added by #453).
2. **Re-runs the judgement under the lock**, via the existing `read_index()`, instead of trusting the earlier unlocked verdict. Taking the lock alone is not sufficient — this call could still block on `SaveLock::acquire` and be granted it *right after* a repairing `save()` releases it, in which case the verdict it's holding is already stale. Re-reading closes that: `save()` cannot be mid-write while the lock is held, so whatever this read sees now cannot change again before the rename that follows it.

If the fresh read still says `Undecodable`, `quarantine()` proceeds exactly as before — still a bare rename, but now called only once the caller has re-confirmed the bytes are corrupt *and* holds the lock that guarantees nothing can replace them before the rename lands. If the fresh read now says `Loaded` (the repair landed in the window), that index is returned directly — the repair survives and the caller sees it immediately instead of falling through to "index absent".

### Lock ordering / deadlock argument

Exactly one lock is ever acquired by this path — `SaveLock` on `<path>.lock` — held for no longer than one `read_index` plus one `rename`, released via `Drop` on every return path. Nothing this function calls acquires a lock of its own, and it never acquires a second `SaveLock` (for this or any other index file) while holding this one. `save()` takes the identical single lock for the identical file. Two call sites, one lock each, no nesting on either side — there is no cycle for a deadlock to form on.

## Why the regression test needed a synchronisation hook

The window between the unlocked read and the (pre-fix) rename is a handful of CPU instructions with no syscall in between — sub-microsecond. A concurrent `save()` (serialise + write + two `fsync`s + rename) cannot land inside that window by scheduling luck. Relying on luck would make the test flaky at best, and at worst silently never hit the window and pass against the unfixed code for the wrong reason — the exact failure mode `regression_406.rs` shipped with.

So `vector_index.rs` gains one small, unconditionally-compiled, env-var-gated rendezvous point (`VectorIndex::test_pause_before_quarantine_lock`), inert everywhere except the one child process this test sets `SPARROWDB_TEST_QUARANTINE_PAUSE_DIR` for. It pauses `quarantine_after_recheck` immediately before `SaveLock::acquire` — precisely the point issue #464 named — until told to resume.

### Interleaving forced by `regression_464_quarantine_race.rs` (two real processes)

| step | opener (child A) | repairer (child B) |
|---|---|---|
| 1 | | loads the index while it is still healthy |
| 2 | | *(parent flips one payload byte, header intact)* |
| 3 | | inserts one more id in memory (the "fix") |
| 4 | calls `load_and_quarantine`; unlocked read sees the corrupt bytes | waits |
| 5 | pauses immediately before `SaveLock::acquire` | waits |
| 6 | *(paused)* | `save()`: acquires the lock, writes, fsyncs, renames, releases |
| 7 | resumes: acquires the now-free lock, re-reads, sees the repaired file | *(exited)* |

Assertions: the opener reports the repaired index (not an error, not absent, not a stale corrupt-count); the good file is still at `path` and still loads; **no** `.corrupt.` artifact exists anywhere; `GraphDb::vector_index_load_failures` reports healthy.

## Pre-fix observation (measured)

Checked out `5f36b02` (the direct parent, confirmed via `git merge-base --is-ancestor`) and applied *only* the new test file, unmodified:

```
$ cargo test -p sparrowdb --test regression_464_quarantine_race -- --nocapture
...
thread 'quarantine_recheck_survives_a_concurrent_repair' panicked at .../regression_464_quarantine_race.rs:106:9:
parent (waiting for opener to pause): .../pause/paused never appeared within 60s
test result: FAILED. 0 passed; 1 failed; 2 ignored
```

That failure is real (the test cannot pass without the fix — the pause hook it depends on does not exist pre-fix), but it doesn't yet exhibit the bug's actual symptom, since the pre-fix binary has no rendezvous point to pause at. To observe the symptom directly, I additionally applied a **diagnostic-only** patch to the *unfixed* `load_and_quarantine` (not part of the shipped fix — mirrors the shipped hook, inserted at the exact call site the issue names, `vector_index.rs:621-647`, immediately before `Self::quarantine(&path)`). With that in place:

```
thread 'quarantine_recheck_survives_a_concurrent_repair' panicked:
assertion `left == right` failed: the opener must see the repaired index ...
  left: "ERROR HNSW index .../hnsw_L_p.bin is corrupt (CRC32C mismatch: header says 0xe45db7bf,
         payload hashes to 0x4523990c); the damaged file was preserved as
         .../hnsw_L_p.bin.corrupt.1785725577128"
 right: "LOADED 6 true"
```

I then loaded the quarantined artifact directly (via `VectorIndex::load`) to confirm what actually got renamed:

```
DIAGNOSTIC: quarantined artifact decodes with len=6 has_vector(999)=true has_vector(0)=true
```

**The file moved into `*.corrupt.*` was the healthy, fully repaired 6-vector index** (5 seeded ids + the repairer's id 999) — exactly issue #464's description: the pair reads as absent, and the quarantine artifact — the only place an operator would look — is perfectly healthy. The error message's stale "CRC32C mismatch" reason (captured before the repair) no longer describes the bytes it was attached to, which is itself part of what makes this failure mode so misleading in production.

## Post-fix observation (measured)

Same interleaving, same test, against this branch:

```
$ cargo test -p sparrowdb --test regression_464_quarantine_race -- --nocapture
test quarantine_recheck_survives_a_concurrent_repair ... ok
test result: ok. 1 passed; 0 failed; 2 ignored
```

Run 15 times back-to-back directly against the compiled test binary (bypassing `cargo test`'s per-invocation overhead) — 15/15 passed, ~20–30ms each, confirming the hook resolves via the rendezvous flags rather than the 60s timeout.

`cargo test -p sparrowdb --test regression_464_quarantine_race -- --ignored --nocapture` confirms both worker tests (`child_opener`, `child_repairer`) return cleanly (not panic) when their env vars are unset, so `cargo test -- --ignored` sweeps stay green.

## Other verification

- `cargo test -p sparrowdb-storage --lib`: 137 passed, 0 failed, 1 ignored.
- `cargo test -p sparrowdb --test vector_index_durability --test vector_index_load_errors --test regression_456_load_is_not_destructive --test regression_452`: 29 passed, 0 failed across all four files.
- `cargo check --workspace --exclude sparrowdb-ruby`: clean.
- `cargo fmt --all -- --check`: clean (after `cargo fmt --all`).
- `cargo clippy -p sparrowdb-storage -p sparrowdb --all-targets --keep-going -- -D warnings`: clean, 0 warnings.
- Built with `CARGO_TARGET_DIR=/Volumes/Dev/target-464`, isolated from sibling worktrees building concurrently.

## Could not verify

- **The full `cargo test -p sparrowdb --tests` suite (167 test binaries)** was still compiling/running in the background when this PR was opened and was not waited on to completion — per this repo's own documented timing (`spa_222_csr_lazy_load` alone runs 5+ minutes; a full run can exceed 60 minutes), finishing it synchronously wasn't practical for this change. The targeted subset above (storage lib + every vector-index/quarantine-adjacent integration test file) is what's actually exercised by this diff; I have not observed unrelated failures in anything I did run, and CI will cover the rest.
- **Real multi-process contention beyond this harness** — the regression test forces a specific, deterministic interleaving via the pause hook. I did not additionally stress-test with many concurrent openers/repairers hammering the same file outside that controlled scenario (regression_452 already covers concurrent `save()`-vs-`save()`; this PR doesn't touch that path).
- **Windows/`LockFileEx` behavior** — `SaveLock` documents Unix `flock(2)` vs Windows `LockFileEx` parity in its existing doc comment; I did not build or test on Windows, only macOS (this machine).
- I did not run the full CI-configured toolchain (this repo's CI pins specific versions, e.g. Ruby 3.3 via `ruby/setup-ruby@v1`, that aren't relevant to this change but I haven't independently reproduced the CI environment).

Closes #464.

### DIFF
--- a/crates/sparrowdb-storage/Cargo.toml
+++ b/crates/sparrowdb-storage/Cargo.toml
@@ -26,6 +26,23 @@ serde             = { workspace = true }
 tempfile  = { workspace = true }
 criterion = { workspace = true }
 
+[features]
+# Off by default, and not enabled by anything else in this workspace's
+# dependency graph. Compiles in VectorIndex::test_pause_before_quarantine_lock
+# — a rendezvous hook the multi-process regression test in
+# crates/sparrowdb/tests/regression_464_quarantine_race.rs uses to force a
+# deterministic interleaving with quarantine_after_recheck(). See that
+# function's doc comment for why an unconditionally-compiled or
+# `debug_assertions`-gated version is not acceptable here: a published,
+# embedded database is linked into other people's binaries, including their
+# own debug builds, and this hook can stall the quarantine path for up to 60
+# seconds and write into an environment-named directory. A non-default
+# feature is never enabled by a downstream consumer's ordinary `cargo build`
+# — debug or release — only by something that explicitly asks for
+# `sparrowdb-storage/test-hooks` by name, which nothing in this workspace's
+# published dependency graph does.
+test-hooks = []
+
 [[bench]]
 name = "storage_benchmarks"
 harness = false
--- a/crates/sparrowdb-storage/src/vector_index.rs
+++ b/crates/sparrowdb-storage/src/vector_index.rs
@@ -726,13 +726,28 @@ impl VectorIndex {
     /// Test-only rendezvous point, exercised by
     /// `regression_464_quarantine_race.rs`.
     ///
-    /// Unconditionally compiled rather than `#[cfg(test)]`: the race it lets
+    /// Unconditionally *present* rather than `#[cfg(test)]`: the race it lets
     /// the test force spans a process boundary, so the test binary links
     /// this crate as an ordinary dependency, not under `cfg(test)` — a
-    /// `#[cfg(test)]` item here would simply not exist in that build. It is
-    /// inert everywhere else: `SPARROWDB_TEST_QUARANTINE_PAUSE_DIR` is set
-    /// nowhere except that one test's child process, so every real caller,
-    /// and every other test, returns from this on the first line.
+    /// `#[cfg(test)]` item here would simply not exist in that build.
+    ///
+    /// It is gated on `debug_assertions` instead, which is the property that
+    /// actually matters: absent from every release build this crate ships,
+    /// present in the debug/dev-profile build the integration test binary
+    /// links (Cargo's default `[profile.test]` inherits `debug-assertions =
+    /// true` from `[profile.dev]`, and this workspace does not override
+    /// that). A `#[cfg(not(debug_assertions))]` build sees the no-op stub
+    /// below, which the optimiser deletes entirely — so this test does not
+    /// run, and must not be relied on, under `cargo test --release`.
+    ///
+    /// This distinction is not cosmetic. Even gated to the corrupt-file path
+    /// and requiring an env var, an unconditionally-compiled version of this
+    /// hook would ship to every consumer of this crate: anything able to set
+    /// `SPARROWDB_TEST_QUARANTINE_PAUSE_DIR` in a release process's
+    /// environment could stall it for up to 60 seconds inside
+    /// `load_and_quarantine` and cause a file write to a directory of its
+    /// choosing. `debug_assertions` removes that surface from what actually
+    /// ships, rather than merely making it hard to trigger.
     ///
     /// # Why this exists at all
     ///
@@ -749,6 +764,7 @@ impl VectorIndex {
     /// here, on request, lets the test hold this function open for as long
     /// as it needs to let a real `save()` complete, then release it, so the
     /// interleaving is forced rather than hoped for.
+    #[cfg(debug_assertions)]
     fn test_pause_before_quarantine_lock() {
         let Ok(dir) = std::env::var("SPARROWDB_TEST_QUARANTINE_PAUSE_DIR") else {
             return;
@@ -766,6 +782,13 @@ impl VectorIndex {
             std::thread::yield_now();
         }
     }
+
+    /// Release-build stand-in for the hook above: always a no-op, and small
+    /// enough that the optimiser removes the call entirely. See the doc
+    /// comment on the `debug_assertions` version for why the two must not be
+    /// merged into one unconditionally-compiled function.
+    #[cfg(not(debug_assertions))]
+    fn test_pause_before_quarantine_lock() {}
 
     fn inconsistent_error(path: &Path, reason: &str) -> std::io::Error {
         std::io::Error::new(
--- a/crates/sparrowdb-storage/src/vector_index.rs
+++ b/crates/sparrowdb-storage/src/vector_index.rs
@@ -731,23 +731,36 @@ impl VectorIndex {
     /// this crate as an ordinary dependency, not under `cfg(test)` — a
     /// `#[cfg(test)]` item here would simply not exist in that build.
     ///
-    /// It is gated on `debug_assertions` instead, which is the property that
-    /// actually matters: absent from every release build this crate ships,
-    /// present in the debug/dev-profile build the integration test binary
-    /// links (Cargo's default `[profile.test]` inherits `debug-assertions =
-    /// true` from `[profile.dev]`, and this workspace does not override
-    /// that). A `#[cfg(not(debug_assertions))]` build sees the no-op stub
-    /// below, which the optimiser deletes entirely — so this test does not
-    /// run, and must not be relied on, under `cargo test --release`.
+    /// It is gated on the non-default `test-hooks` Cargo feature instead —
+    /// deliberately not `debug_assertions`, which was this crate's first
+    /// attempt and has a real gap: `debug_assertions` is on for *every*
+    /// debug build, including a downstream consumer's own `cargo build` of
+    /// their application that happens to link this crate. This is an
+    /// embedded, published database other people's binaries link directly,
+    /// so "off in release" is not the same bar as "off unless something
+    /// explicitly asks for it by name". A non-default feature clears that
+    /// bar: nobody gets `test-hooks` — debug or release — without an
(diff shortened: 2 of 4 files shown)


ryaker:
@coderabbitai review

This closes a race in which a concurrent repair could be silently destroyed. Specific areas where I want scrutiny:

1. **The re-judge under the lock.** `quarantine_after_recheck` acquires `SaveLock` and then re-runs `read_index` before renaming, rather than comparing a fingerprint captured earlier. Is the critical section genuinely closed — i.e. can anything still replace the file between that second read and the rename?

2. **Lock discipline.** Is `SaveLock` released on *every* return path, including the `?` on acquire and each early return in the match? Is there any path where `quarantine_after_recheck` could be reached while a `SaveLock` for the same file is already held higher in the stack (self-deadlock, since this is `flock`-based and not reentrant)?

3. **The test hook.** `test_pause_before_quarantine_lock` is unconditionally compiled and gated only on the `SPARROWDB_TEST_QUARANTINE_PAUSE_DIR` env var, so it exists in release builds. Assessment of that as a production surface would be welcome — it can stall for up to 60s and writes a file into an env-supplied directory.

4. **Behaviour change**: when the fresh read returns `Loaded`, the index is now handed back to the caller instead of the pair reading as absent. Is that safe for every caller of `load_and_quarantine`?

ryaker:
@coderabbitai review

Issue #467, opened 2026-08-09
Pattern-prop filter fails OPEN on Null: MATCH (n:L {p: <null expr>}) silently matches every node


ryaker:
## The defect

`matches_prop_filter_static` (`crates/sparrowdb-execution/src/engine/mod.rs:1178`) ends its match arms with:

```rust
Value::Null => true, // null filter passes (param-like behaviour)
```

Any pattern-property filter whose value expression evaluates to `Value::Null` is **skipped entirely**, so the pattern degrades to "every node of this label." On a read that returns too many rows; on a mutation it writes to every node of the label.

A filter that cannot be evaluated should fail **closed** (match nothing) or raise. Failing open turns an unresolvable expression into a silent full-label operation.

## Confirmed reachable today

`Expr::Var` that is not present in the params map evaluates to `Null`, and `resolve_pattern_props` / `row_expr_to_literal` only resolve `Expr::PropAccess`. So on the SPA-415 branch (#411):

```cypher
UNWIND $ids AS row MATCH (n:Item {id: row}) SET n.val = 5
```

sets `val = 5` on **every** `:Item`, not the one named. Reproduced with 3 nodes and 1 named id — all three came back `Int64(5)`. Test written and currently `#[ignore]`d as `unwind_pattern_prop_bare_alias_var_must_not_match_every_node`; it fails when run with `--ignored`, which is what proves the bug is real.

This is an easy mistake to make: `UNWIND list AS x MATCH (n {prop: x})` is the standard idiom in other Cypher engines when `x` is a scalar. Here `x` is always a Map, so the correct form silently requires `.field`.

## Why this is filed separately from #411

The `unwrap_or`-style fail-open in `matches_prop_filter_static` is **not** introduced by #411 — #411 merely supplies a reachable path to it. The audit worth doing is whether other unbound or unresolvable expressions reach a pattern-prop filter: an unbound correlated variable in a plain `MATCH`, a `$param` absent from the map, a function call returning Null.

Same shape as the defects this repo keeps finding: a value that means "I could not determine this" being treated as "no constraint." Compare #417's `Token::Ident("KEYWORD")` guards that never fire, and the `next_label_id_opt.unwrap_or(0)` case where label id 0 is a real label rather than a sentinel.

## Suggested fix

[...]

PR #470, opened 2026-08-09
feat(cypher): UNWIND MATCH SET/DELETE + DELETE RETURN (SPA-415) — rebase of #411 with tests (closes #468)

### LINKED ISSUE #468: PR #411 regresses UNWIND … MATCH … RETURN: MATCH silently dropped when both leading_unwind and leading_match are set
## Summary

PR #411 (UNWIND MATCH SET/DELETE + DELETE RETURN, SPA-415) **breaks the previously-working `UNWIND … MATCH … RETURN` read pipeline**. This blocks that PR.

## Verified independently

| tree | result |
|---|---|
| `origin/main` @ `6254f91` | `spa_237_unwind_match`: **6 passed, 0 failed** |
| `pr411-rebase` @ `62a6931` | `spa_237_unwind_match`: **4 passed, 2 failed** |

Failures:
- `unwind_list_then_match_filters_by_id` — `Expected Alpha in results, got ["null", "null"]`
- `unwind_list_element_with_no_match_is_skipped` — expected 1 row, got 2

Bisected to `9600c0e`, the PR's first commit; still present at HEAD.

## Root cause

`execute_pipeline_inner` (`crates/sparrowdb-execution/src/engine/scan.rs`) selects the leading clause with an `if / else if` chain that tests **`leading_unwind` before `leading_match`**:

```rust
} else if let Some((expr, alias)) = &p.leading_unwind {   // ~line 47
    ...
} else if let Some(ref patterns) = p.leading_match {      // ~line 57
```

`parse_unwind`'s new dispatch (`parser.rs` ~1657) pre-consumes `MATCH` and calls `parse_pattern_list()` itself to decide mutate-vs-read, then for the read branch calls:

```rust
parse_pipeline_continuation(Some(patterns), None, Some((expr, alias)), vec![])
```

passing the patterns as `leading_match`. That makes **both** fields `Some` — a combination no other code path produces — so the `leading_match` arm is unreachable and the MATCH is dropped silently. Every UNWIND element becomes a row with no `n` binding, so `n.name` projects as NULL and the row count no longer reflects the filter.

Before #411, `MATCH` after `UNWIND` was left for `parse_pipeline_continuation`'s own loop to consume as a `PipelineStage::Match`, executed in sequence after the UNWIND stage. The early consumption bypasses that.

## Why it wasn't caught

The `if/else if` makes the dropped clause invisible — no error, no warning, just NULLs and a wrong row count. The PR shipped with **zero tests**; `spa_237_unwind_match` already covered this exact case and would have caught it on the first run.

## Fix direction

Either stop setting `leading_match` in the read branch (let the continuation loop consume the MATCH as a stage, as before), or make `execute_pipeline_inner` handle the both-set case explicitly instead of letting an `else if` silently discard one. The first is closer to the pre-#411 behaviour and does not change a shared execution path.

Whatever the fix, `spa_237_unwind_match` must pass, and the new `spa_415_unwind_match_mutate` suite (15 tests) must stay green.

Related: #467 (pattern-prop filters fail open on Null — a second defect on the same branch).

ryaker:
Rebase of **#411 by @evuk79** onto current `main`, with the missing test coverage and two regression fixes. Their four commits are preserved with original authorship; #411 should be closed in favour of this once merged, with credit to them.

## What #411 delivered

`UNWIND $list AS x MATCH (n {…}) SET/DELETE …` plus `DELETE … RETURN count()`. The params plumbing is correct — `params` is threaded through `execute_unwind_mutate_inner`, `engine.with_params()`, and `resolve_set_value` (commit `8a08721`), which satisfies all five high-priority findings from the June review. Those threads were never resolved on #411, so its "6 unresolved" count was bookkeeping, not outstanding work.

## What this branch adds

**Tests (`62a6931`).** #411 shipped 576 lines across `ast.rs`/`binder.rs`/`parser.rs`/`engine/mod.rs`/`db.rs` with **zero tests**. New `crates/sparrowdb/tests/spa_415_unwind_match_mutate.rs`, 16 tests, e2e against a real `GraphDb` on a tempdir:

- literal-list and `$param` paths, with a control node that must stay untouched
- `$param` resolving in **both** the WHERE clause and a SET value
- `SET n.prop = row.field` with a different value per row (guards `aa29e59`)
- DELETE, DETACH DELETE, `DELETE … RETURN count()` accumulated across rows
- empty list, no-match element, non-`count()` RETURN limitation
- **durability**: mutate → drop `GraphDb` → reopen → assert on disk
- the documented same-snapshot visibility semantic, asserted so it fails if that semantic ever changes

**Regression fix (`8995ab6`, closes #468).** #411 broke `UNWIND … MATCH … RETURN`, which worked on `6254f91`. Two distinct faults in the new dispatch:

1. Patterns were passed as `leading_match` while `leading_unwind` was also set. `execute_pipeline_inner` picks the leading clause with an `if / else if` chain testing `leading_unwind` first, so the MATCH arm was unreachable and the pattern was dropped silently — `spa_237_unwind_match` returned `["null", "null"]`. Fixed by handing the MATCH over as a `PipelineStage::Match`, the shape `parse_pipeline_continuation`'s own loop produced pre-SPA-415, rather than editing the shared `if/else if` and changing behaviour for every pipeline in the engine.
2. Dispatching on `Token::Where` routed every *read* carrying a predicate into the mutation tail, which rejected it. `UNWIND ['a','b'] AS x MATCH (n:Item {id: x}) WHERE n.v > 3 RETURN n.id` returns `[["a"]]` on main and errored on the branch. WHERE is legal in both shapes, so it is now parsed once up front and the read/mutate decision is made on SET/DELETE/DETACH.

Both regression tests were confirmed to **fail against the pre-fix parser** (`62a6931`) with those exact symptoms.

## Results

| suite | result |
|---|---|
| `spa_415_unwind_match_mutate` | 16 passed, 1 ignored (#467) |
| `spa_237_unwind_match` | 6 passed — back to parity with main |
| `spa_155_unwind_param` | 4 passed |
| `uc7_unwind` | 12 passed |
| `gap_10_parameterized_queries` | 12 passed |
| `sparrowdb-cypher` | all passed |

`cargo fmt` and `cargo clippy -p sparrowdb -p sparrowdb-cypher --all-targets --keep-going` clean.

## Known issue, deliberately not fixed here

The one ignored test documents **#467**: `matches_prop_filter_static` treats a `Null` filter as always-passing, so `MATCH (n:Item {id: row})` with a bare alias (rather than `row.id`) matches **every** node of the label — on a `SET`, it writes to all of them. The fail-open is pre-existing; this feature merely supplies a reachable path to it. Filed separately so it gets the wider audit it needs rather than a narrow patch here.

Closes #468. Supersedes #411.

### DIFF
--- a/crates/sparrowdb-cypher/src/ast.rs
+++ b/crates/sparrowdb-cypher/src/ast.rs
@@ -327,6 +327,34 @@ pub struct MatchMutateStatement {
     /// For SET, there may be multiple items (comma-separated).
     /// For DELETE, this always contains exactly one `Mutation::Delete`.
     pub mutations: Vec<Mutation>,
+    /// Optional RETURN clause after mutations (e.g. `DELETE n RETURN count(n)`).
+    pub return_clause: Option<ReturnClause>,
+}
+
+/// UNWIND … MATCH … SET/DELETE statement (SPA-415).
+///
+/// Iterates a list expression, binding each element to `alias`, then for
+/// each element executes a MATCH + mutation (SET or DELETE).
+///
+/// ```cypher
+/// UNWIND $updates AS row
+/// MATCH (n:Memory {id: row.id})
+/// SET n.score = row.score
+/// ```
+#[derive(Debug, Clone, PartialEq)]
+pub struct UnwindMatchMutateStatement {
+    /// The list expression to unwind (e.g. `$updates`).
+    pub expr: Expr,
+    /// Variable bound to each list element (e.g. `row`).
+    pub alias: String,
+    /// The MATCH patterns whose nodes are mutated.
+    pub match_patterns: Vec<PathPattern>,
+    /// Optional WHERE predicate on the MATCH.
+    pub where_clause: Option<Expr>,
+    /// The mutations to apply (SET or DELETE).
+    pub mutations: Vec<Mutation>,
+    /// Optional RETURN clause after mutations.
+    pub return_clause: Option<ReturnClause>,
 }
 
 /// A single projection in a WITH clause: `expr AS alias`.
@@ -501,6 +529,8 @@ pub enum Statement {
     /// MATCH … MERGE (a)-[r:TYPE]->(b) — find-or-create a relationship (SPA-233).
     MatchMergeRel(MatchMergeRelStatement),
     MatchMutate(MatchMutateStatement),
+    /// UNWIND … MATCH … SET/DELETE — batched mutations via list parameter (SPA-415).
+    UnwindMatchMutate(UnwindMatchMutateStatement),
     OptionalMatch(OptionalMatchStatement),
     MatchOptionalMatch(MatchOptionalMatchStatement),
     Union(UnionStatement),
--- a/crates/sparrowdb-cypher/src/binder.rs
+++ b/crates/sparrowdb-cypher/src/binder.rs
@@ -14,8 +14,8 @@ use sparrowdb_catalog::catalog::Catalog;
 use sparrowdb_common::Result;
 
 use crate::ast::{
-    MatchMergeRelStatement, MatchOptionalMatchStatement, MatchStatement,
-    MatchWithStatement, PathPattern, Statement, UnwindMatchMutateStatement,
+    MatchMergeRelStatement, MatchOptionalMatchStatement, MatchStatement, MatchWithStatement,
+    PathPattern, Statement, UnwindMatchMutateStatement,
 };
 
 /// A bound statement — the AST annotated with resolved catalog IDs.
--- a/crates/sparrowdb-cypher/src/binder.rs
+++ b/crates/sparrowdb-cypher/src/binder.rs
@@ -14,8 +14,8 @@ use sparrowdb_catalog::catalog::Catalog;
 use sparrowdb_common::Result;
 
 use crate::ast::{
-    MatchMergeRelStatement, MatchMutateStatement, MatchOptionalMatchStatement, MatchStatement,
-    MatchWithStatement, PathPattern, Statement,
+    MatchMergeRelStatement, MatchOptionalMatchStatement, MatchStatement,
+    MatchWithStatement, PathPattern, Statement, UnwindMatchMutateStatement,
 };
 
 /// A bound statement — the AST annotated with resolved catalog IDs.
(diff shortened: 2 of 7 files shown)


ryaker:
@coderabbitai full review

Please pay particular attention to the parser dispatch in `parse_unwind` (`crates/sparrowdb-cypher/src/parser.rs`). It pre-consumes MATCH and now WHERE in order to look ahead for SET/DELETE, then hands control to either `parse_unwind_match_mutate_tail` or `parse_pipeline_continuation`. Two regressions already came out of that lookahead (#468) — a pattern passed as `leading_match` while `leading_unwind` was also set, and a WHERE on a read routed into the mutation tail. I would like a check on whether any other token sequence reachable from that dispatch produces a shape the executor mishandles.

Also worth your attention: `unwind_rows_do_not_see_earlier_rows_mutations_within_same_statement` pins a *documented* semantic (each UNWIND row scans the same committed snapshot) rather than a bug. If you think that semantic is wrong for Cypher, that's a more useful finding than anything about the test itself.

ryaker:
Composition with #471 verified, plus a full workspace suite on the combination.

#471 fixes #467, the fail-open pattern-property filter. This PR carries the `#[ignore]`d test that demonstrates #467 through the UNWIND path — `unwind_pattern_prop_bare_alias_var_must_not_match_every_node`. Merged the two branches in a scratch worktree (auto-merged; only `engine/mod.rs` and `parser.rs` were touched, no overlap) and that test **passes**. So whichever of the two lands second should un-ignore it.

Full `cargo test --workspace --exclude sparrowdb-ruby --no-fail-fast` on the composed tree:

```
223 binaries, 1533 passed, 0 failed, 21 ignored
```

Reconciles exactly against the baseline — main is 221 binaries / 1512 passed / 20 ignored, and 221 + `spa_415_unwind_match_mutate` + `regression_467` = 223; 1512 + 16 + 5 = 1533. Every delta accounted for, nothing silently dropped. The one added ignore is the test above.

Worth noting the two PRs are genuinely independent: this one is a parser-dispatch fix, #471 is an engine filter fix. They compose because neither touches the other's surface — verified rather than assumed, since two individually-correct PRs failing to compose has bitten this repo before.

PR #471, opened 2026-08-09
fix(engine): pattern-property filters fail closed on unresolvable expressions (closes #467)

ryaker:
## Summary

`matches_prop_filter_static` (`crates/sparrowdb-execution/src/engine/mod.rs`) ended its match arms with:

```rust
Value::Null => true, // null filter passes (param-like behaviour)
```

Every leaf of a pattern-property filter's value expression that couldn't be resolved — a `$param` never supplied, a bare variable not in scope, a property access on an unbound var — falls back to `Value::Null` via `eval_expr`. That `Null` was then treated as an automatic pass, so `MATCH (n:Label {prop: <unresolvable>})` silently degraded into "every live node of `Label`". On a read this over-returns rows; on `SET`/`DELETE` it mutates or removes nodes the query never named.

This fix separates the two conditions that were being conflated:

- **Genuinely bound to null** (an explicit `null` literal, or a `$param` the caller deliberately supplied as null) — now matches only a node whose own property is likewise absent, mirroring the `Null == Null` convention `values_equal` already uses for WHERE-clause equality elsewhere in the same file. It no longer matches every node regardless of stored value.
- **Unresolvable** (missing `$param`, unbound variable, property access on a var not in scope) — now fails closed unconditionally via a new `is_filter_expr_resolvable` check, run *before* the filter value is ever evaluated.

`matches_prop_filter_static` is the single implementation behind every pattern-property filter call site in the engine — scan, hop, path, mutation, and the chunked pipeline all funnel through it or its `matches_prop_filter` wrapper — so this one change closes the hole everywhere it was reachable.

## Audit

| Null-producing path | Reachable today | Confirmed | Query |
|---|---|---|---|
| `$param` absent from params map | Yes | Ran it | `MATCH (n:Item {id: $missing}) RETURN n.id` (missing) |
| `$param` absent, mutation (SET) | Yes | Ran it | `MATCH (n:Item {id: $missing}) SET n.val = 99` — wrote to every node pre-fix |
| Bare unbound `Expr::Var` in sibling pattern-prop position | Yes | Ran it | `MATCH (a:A), (n:Item {id: a}) RETURN n.id` |
| `Expr::PropAccess` cross-pattern correlation | Yes | Ran it | `MATCH (a:A), (n:Item {id: a.tag}) RETURN n.id` |
| Genuinely-null-bound `$param` (control case, not a bug) | Yes | Ran it | pins the chosen semantic in `regression_467.rs` |

All five call sites of `matches_prop_filter_static`/`matches_prop_filter` (mutation.rs, pipeline_exec.rs, scan.rs, expr.rs ×3) pass only `dollar_params()` — never row-scope bindings — so a bare `Expr::Var`/`Expr::PropAccess` referencing an outer pattern variable is *never* resolvable through this function today, regardless of query shape. That's why the fail-closed check is unconditional for those expression kinds rather than attempting (and only partially succeeding at) resolution.

**Sibling path checked and found safe in the failure direction:** `matches_prop_filter_with_binding` (`scan.rs:635`, used by the pipeline `WITH … MATCH` re-traversal stage) has the same conflation for its `Expr::Var` arm, but its match logic (`(None, Value::Null) => true, _ => false`) already fails toward *too few* matches rather than *every* node — confirmed by test, not fixed here (documented as a lower-severity gap, not the #467 vulnerability).

**Related, out of scope:** `value_to_store_value` (`engine/mod.rs:1536`) silently writes `Value::Null` as `StoreValue::Int64(0)` on the `CREATE` prop-write path — a different bug shape (wrong value written vs. wrong nodes matched), not touched here.

## Testing

`crates/sparrowdb/tests/regression_467.rs` — 5 e2e tests against a real `GraphDb` on `tempfile::TempDir`:
1. Read with missing `$param` → empty result, not all 3 nodes.
2. **SET with missing `$param`** → untouched control node, catches the data-corruption case.
3. Bare unbound `Expr::Var` in a sibling pattern's prop filter → empty result.
4. Genuinely-null-bound `$param` vs. a node with an absent property → matches only that node.
5. Genuinely-null-bound `$param` where every candidate has the property set → matches none.

All 5 confirmed to **fail against the pre-fix commit** (`6254f91`, stashing just the engine change) with output matching the reported symptom exactly, e.g.:
```
SET with an unresolvable $param must not write to any node (data-corruption case),
got [[String("a"), Int64(99)], [String("b"), Int64(99)], [String("c"), Int64(99)]]
```

## Blast radius

- `cargo fmt` clean, `cargo clippy -p sparrowdb -p sparrowdb-execution --all-targets --keep-going` clean.
- Targeted suites (all pass, no behaviour changes): `gap_10_parameterized_queries`, `spa_289_multi_label`, `spa157_cypher_mutations`, `spa163_164_read_path`, `uc1_social_graph`, `uc7_unwind`, `spa_299_chunked_pipeline`, `spa_299_phase2/3/4_parity` — 89 tests, 2 pre-existing unrelated `#[ignore]`s.
- Full `cargo test --workspace --exclude sparrowdb-ruby --no-fail-fast` run in progress; will report results as a follow-up comment.
- No existing test in the repo depended on the old "null filter matches every node" behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pattern-property filters now safely return no matches when required parameters, expressions, or functions cannot be resolved.
  * Explicit `null` filters now match only nodes where the specified property is absent.
  * Prevented unresolved filters from unintentionally matching all nodes or modifying data.
  * Fully resolvable conditional expressions and supported functions continue to match as expected.

* **Tests**
  * Added regression coverage for missing parameters, variables, conditional expressions, unknown functions, `null` values, and node mutation behavior.

### DIFF
--- a/crates/sparrowdb-execution/src/engine/mod.rs
+++ b/crates/sparrowdb-execution/src/engine/mod.rs
@@ -1155,6 +1155,19 @@ mod subquery;
 /// query referenced a parameter the caller never supplied, not that the
 /// caller deliberately bound it to null.
 fn is_filter_expr_resolvable(expr: &Expr, params: &HashMap<String, Value>) -> bool {
+    is_filter_expr_resolvable_scoped(expr, params, &[])
+}
+
+/// As [`is_filter_expr_resolvable`], but with a set of locally-bound variable
+/// names — the loop variable of a list predicate (`ANY(x IN … WHERE …)`) is
+/// bound by the comprehension itself, not by the params map, so a bare
+/// `Expr::Var("x")` inside the predicate is legitimately resolvable.
+fn is_filter_expr_resolvable_scoped(
+    expr: &Expr,
+    params: &HashMap<String, Value>,
+    locals: &[&str],
+) -> bool {
+    let rec = |e: &Expr| is_filter_expr_resolvable_scoped(e, params, locals);
     match expr {
         // A literal `$name` — resolvable only if the caller actually supplied it.
         Expr::Literal(Literal::Param(p)) => params.contains_key(&format!("${p}")),
@@ -1175,23 +1188,16 @@ fn is_filter_expr_resolvable(expr: &Expr, params: &HashMap<String, Value>) -> bo
         // duplicated list of known names, which would silently drift as
         // functions are added.
         Expr::FnCall { name, args } => {
-            args.iter().all(|a| is_filter_expr_resolvable(a, params)) && {
+            args.iter().all(&rec) && {
                 let evaluated: Vec<Value> = args.iter().map(|a| eval_expr(a, params)).collect();
                 crate::functions::dispatch_function(name, evaluated).is_ok()
             }
         }
-        Expr::BinOp { left, right, .. } => {
-            is_filter_expr_resolvable(left, params) && is_filter_expr_resolvable(right, params)
-        }
-        Expr::List(items) => items.iter().all(|e| is_filter_expr_resolvable(e, params)),
-        Expr::InList { expr, list, .. } => {
-            is_filter_expr_resolvable(expr, params)
-                && list.iter().all(|e| is_filter_expr_resolvable(e, params))
-        }
-        Expr::Not(e) | Expr::IsNull(e) | Expr::IsNotNull(e) => is_filter_expr_resolvable(e, params),
-        Expr::And(a, b) | Expr::Or(a, b) => {
-            is_filter_expr_resolvable(a, params) && is_filter_expr_resolvable(b, params)
-        }
+        Expr::BinOp { left, right, .. } => rec(left) && rec(right),
+        Expr::List(items) => items.iter().all(&rec),
+        Expr::InList { expr, list, .. } => rec(expr) && list.iter().all(&rec),
+        Expr::Not(e) | Expr::IsNull(e) | Expr::IsNotNull(e) => rec(e),
+        Expr::And(a, b) | Expr::Or(a, b) => rec(a) && rec(b),
         // `CASE WHEN <cond> THEN <val> … [ELSE <val>]` is resolvable iff every
         // condition, every branch value, and the ELSE are.  This one is NOT
         // merely conservative: `MATCH (n:Item {id: CASE WHEN true THEN 1 ELSE 2 END})`
@@ -1201,12 +1207,26 @@ fn is_filter_expr_resolvable(expr: &Expr, params: &HashMap<String, Value>) -> bo
             branches,
             else_expr,
         } => {
-            branches.iter().all(|(cond, val)| {
-                is_filter_expr_resolvable(cond, params) && is_filter_expr_resolvable(val, params)
-            }) && else_expr
-                .as_ref()
-                .is_none_or(|e| is_filter_expr_resolvable(e, params))
+            branches.iter().all(|(cond, val)| rec(cond) && rec(val))
+                && else_expr.as_ref().is_none_or(|e| rec(e))
         }
+        // `ANY/ALL/NONE/SINGLE (x IN <list> WHERE <pred>)` binds `x` itself, so
+        // the predicate is resolvable even though it names a variable absent
+        // from the params map. Omitting this arm dropped the whole expression
+        // into the catch-all below, silently discarding a legitimate row — the
+        // same symptom-free under-match as the CASE WHEN gap above.
+        Expr::ListPredicate {
+            variable,
+            list_expr,
+            predicate,
+            ..
+        } => {
+            let mut inner: Vec<&str> = locals.to_vec();
+            inner.push(variable.as_str());
+            rec(list_expr) && is_filter_expr_resolvable_scoped(predicate, params, &inner)
+        }
+        // A comprehension-bound loop variable is resolvable inside its own body.
+        Expr::Var(v) if locals.contains(&v.as_str()) => true,
         // A bare variable or property access can never be resolved from a
         // params-only map — see the doc comment above.  Everything else
         // (EXISTS, shortestPath, CountStar, list predicates, …) is not
--- a/crates/sparrowdb-execution/src/engine/mod.rs
+++ b/crates/sparrowdb-execution/src/engine/mod.rs
@@ -1162,7 +1162,24 @@ fn is_filter_expr_resolvable(expr: &Expr, params: &HashMap<String, Value>) -> bo
         // are always resolvable.
         Expr::Literal(_) => true,
         // Function calls / arithmetic are resolvable iff every argument is.
-        Expr::FnCall { args, .. } => args.iter().all(|a| is_filter_expr_resolvable(a, params)),
+        // A function call is resolvable iff every argument is AND the dispatcher
+        // actually accepts the call. Checking the arguments alone is not enough:
+        // the parser accepts any identifier as a function name, and
+        // `eval_expr` turns a dispatcher error into `Value::Null`
+        // (`dispatch_function(...).unwrap_or(Value::Null)`). That Null then hits
+        // the `Value::Null => stored_val.is_none()` arm below, so
+        // `MATCH (n:Item {id: bogus_fn(1)})` matched every node *lacking* `id`
+        // instead of matching nothing.
+        //
+        // The dispatcher itself is the source of truth here rather than a
+        // duplicated list of known names, which would silently drift as
+        // functions are added.
+        Expr::FnCall { name, args } => {
+            args.iter().all(|a| is_filter_expr_resolvable(a, params)) && {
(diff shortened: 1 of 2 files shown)


ryaker:
Composition verified against #470 — this is the check that matters for these two, and it passes.

#470 (SPA-415 UNWIND MATCH SET/DELETE, rebase of #411) carries the ignored test `unwind_pattern_prop_bare_alias_var_must_not_match_every_node`, which demonstrates #467 through the UNWIND path. You correctly noted you couldn't test that composition because the feature isn't on `main` yet. I built it:

```
git worktree add … -b compose-467-470 origin/feat/spa-415-unwind-match-mutate
git merge origin/fix/issue-467-filter-fails-open     # auto-merged, only engine/mod.rs overlapped
cargo test -p sparrowdb --test spa_415_unwind_match_mutate -- --ignored
  test unwind_pattern_prop_bare_alias_var_must_not_match_every_node ... ok
```

So this fix does resolve the UNWIND case, and whichever of the two merges second should un-ignore that test.

Composed-tree results, no interaction failures:

| suite | result |
|---|---|
| `regression_467` | 5 passed |
| `spa_415_unwind_match_mutate` | 16 passed, 1 ignored (the above — un-ignore on second merge) |
| `spa_237_unwind_match` | 6 passed |
| `spa_155_unwind_param` | 4 passed |
| `uc7_unwind` | 12 passed |
| `gap_10_parameterized_queries` | 12 passed |

The two branches touch `engine/mod.rs` and `parser.rs` respectively and merged cleanly.

On the two gaps you flagged rather than pulled into scope — both were the right call. `matches_prop_filter_with_binding` failing toward *under*-match is a materially different severity from failing toward writing every node, and mixing an under-match fix into a data-corruption fix would have obscured both. Same for the `value_to_store_value` null→`Int64(0)` write, which is a wrong-value bug rather than a wrong-rows bug. I'll file both as follow-ups.

ryaker:
Full workspace suite on the composed tree (this PR merged into #470's branch) — `223 binaries, 1533 passed, 0 failed, 21 ignored`, reconciling exactly against main's 221/1512/20 plus the two new test binaries. No interaction failures anywhere in the suite, not just in the targeted files.

ryaker:
@coderabbitai full review

Requesting alone — CodeRabbit's included-review quota is roughly one per hour repo-wide, so batching requests across PRs silently starves all but one.

This is the highest-priority PR in the current stack. The bug it fixes is live in shipped 0.1.22: `MATCH (n:Item {id: $missing}) SET n.val = 99` with an absent $param writes to **every** node of the label, because `matches_prop_filter_static` treated a `Value::Null` filter as always-passing.

Please pay particular attention to **over-strictness** in the new `is_filter_expr_resolvable`. Failing closed on something genuinely resolvable would silently drop legitimate matches — a data-visibility bug that is harder to notice than the over-match it replaces. Specifically: a `$param` legitimately bound to null, a property genuinely absent from a node, and function calls whose arguments are all resolvable.

Second area worth your attention: whether the chunked pipeline (`can_use_chunked_pipeline`, live for essentially every `db.execute()`) reaches the same filter logic or bypasses it. A fix covering only the non-chunked path would look complete and not be.

ryaker:
Adversarial review (fresh context, diff-first) found one regression introduced by this PR and one false claim in its description. Both are now addressed in `45f12aa`.

**Regression — fixed.** `is_filter_expr_resolvable` had no `Expr::CaseWhen` arm, so a fully-literal CASE fell into the catch-all beside EXISTS and shortestPath:

```
MATCH (n:Item {id: CASE WHEN true THEN 1 ELSE 2 END}) RETURN n.id
  6254f91 (pre-guard): [[1]]   correct
  0384ff0 (this PR):   []      row silently dropped
```

I reproduced both sides before fixing. CASE is now resolvable iff every condition, branch value, and the ELSE are — so a CASE carrying an unresolvable sub-expression still fails closed, and both directions are tested. The positive test fails against `0384ff0` with the row-dropping symptom.

This is the failure mode worth taking seriously here: an **under-match is harder to notice than the over-match this guard exists to prevent** — no error, just a missing row. The catch-all now carries a note that adding a variant to it is only correct when that variant genuinely cannot be evaluated from literals and params.

**False claim in the PR description — corrected.** The body said `matches_prop_filter_with_binding` was "checked and found safe in the failure direction." It is not. When the filtered property has *never been set by any node of the label*, that column is never fetched, `stored_val` is genuinely `None`, and `(None, Value::Null) => true` fires for every row — reproducing #467's exact match-everything shape. The earlier test only looked safe because it used a property that existed on another node, where `get_node_raw` zero-sentinels absent slots to `Some(Int64(0))`.

Bounded to reads: `PipelineStage` rejects `SET` after a pipeline `MATCH`, so it over-returns but cannot mutate. It is pre-existing (`git diff origin/main..HEAD -- scan.rs` → 0 lines) and now escalated on #472 with the repro, so this PR is an **incomplete** fix for #467's family rather than a complete one — I have not edited the claim out quietly.

**Third finding, logged not fixed:** a genuinely-null-bound `$param` behaves differently on read vs mutation, because `mutation.rs::node_matches_prop_filter`, `expr.rs::find_node_by_props`, and the EXISTS-pattern check source props via plain `get_node_raw`, which zero-sentinels rather than yielding `None`. It fails *safe* — a null-bound filter matches nothing instead of everything — so it is a fast-follow, not a blocker. Filing separately.

regression_467 7/7, gap_10_parameterized_queries 12/12, spa_299_chunked_pipeline 23/23, spa_131_optional_match, spa_237_unwind_match all green; clippy clean.

ryaker:
Fixed in \`15dc336\` — confirmed real before fixing, rather than taken on the description.

\`\`\`
// node a has id, node b does not
MATCH (n:Item {id: bogus_fn(1)}) RETURN n.other
  before: [[20]]   <- node b, the one with no `id` at all
  after:  []
\`\`\`

You had the mechanism exactly right: the parser accepts any identifier as a function name, `eval_expr` does `dispatch_function(...).unwrap_or(Value::Null)`, and that Null reached the new `Value::Null => stored_val.is_none()` arm.

`is_filter_expr_resolvable` now asks **the dispatcher itself** whether the call succeeds, rather than consulting a duplicated list of known names — a list would drift silently as functions are added, which is the same duplication that produced the `matches_prop_filter_with_binding` gap now tracked in #472.

Two tests, both directions: the negative fails against `45f12aa` with the row above, and `known_function_in_prop_filter_still_matches` (`abs(-1)` → the `id=1` node) guards against satisfying the check by rejecting every call. That over-strict direction matters here — a separate reviewer caught this PR silently dropping a legitimate row for a fully-literal `CASE WHEN`, fixed in `45f12aa`.

regression_467 9/9, gap_10 12/12, spa_299_chunked_pipeline 23/23, spa_140_143_functions 27/27, spa_237_unwind_match 23/23; clippy clean.

ryaker:
Third instance of the same class found and fixed in \`b8a2ebc\`. Confirmed before fixing:

\`\`\`
MATCH (n:Item {flag: ANY(x IN [1,2,3] WHERE x > 1)}) RETURN n.tag
  before: []          legitimate row silently dropped
  after:  [["yes"]]
\`\`\`

`Expr::ListPredicate` had no arm and fell into the catch-all. The wrinkle is that the comprehension **binds its own loop variable**, so the predicate is resolvable while naming a variable absent from the params map — which is why a naive recursion could not accept it. `is_filter_expr_resolvable` now delegates to a scoped variant carrying locally-bound names, with every arm recursing through it. A bare `Expr::Var` resolves only inside the comprehension that binds it, so `MATCH (n:Item {flag: x})` still fails closed, and unbound parts fail closed in every position — predicate body, the list itself, and leakage outside the comprehension are all tested.

**That is three now** — CASE WHEN (`45f12aa`), undispatchable functions (`15dc336`), list predicates (`b8a2ebc`). The catch-all is a silent-under-match generator by construction: every statically-evaluable variant that lands there drops rows with no error, which is strictly harder to notice than the over-match this guard exists to prevent. I re-checked the remaining rejections against `eval_expr` and they are correct — `PropAccess`/`Var` need genuinely-absent row context, and `NotExists`/`ExistsSubquery`/`ShortestPath`/`CountStar` need graph traversal or aggregation.

The perf finding (dispatcher called twice per candidate node, quadratic in nesting depth) is real and measured — 10 dispatches for 5 rows at depth 1, 25 at depth 2 — but not correctness-affecting: no dispatchable function's `Ok`/`Err` outcome depends on non-deterministic state, so the throwaway call cannot flip a fail-closed decision. Filed as a follow-up rather than restructuring the evaluation path inside this PR, since that risks reintroducing the exact under-match class above.

regression_467 11/11, gap_10 12/12, spa_299_chunked_pipeline 23/23, spa_140_143_functions 27/27, spa_131_optional_match 10/10, spa_237_unwind_match 6/6; clippy clean.

Issue #487, opened 2026-08-14
execute_two_hop ignores first-hop direction: (a)<-[:R]-(m)-[:R]->(b) silently traverses as if the first hop were outbound


ryaker:
## Found while investigating #486 (edge direction inverted)

`execute_two_hop` (crates/sparrowdb-execution/src/engine/hop.rs, starting line 829) has explicit, tested handling for `rels[1].dir == Incoming` (the second hop — this is the SPA-201 / #294 fix, `second_hop_incoming`). But **`rels[0].dir` (the first hop) is never read anywhere in the function.** The traversal always walks the first hop forward via `hop1_csr` / `csr_neighbors`, regardless of whether the pattern wrote `-[:R]->` or `<-[:R]-` for that hop.

So a pattern like:

```
MATCH (a:Person)<-[:R]-(m:Person)-[:R]->(b:Person) RETURN m.name
```

is silently executed as if it had been written `(a)-[:R]->(m)-[:R]->(b)` — same failure mode as #486, just one function over: wrong rows returned with no error, rather than a `<-` on the first hop being rejected or handled correctly.

## Reproduction sketch

```
CREATE (:P {name:'a'})
CREATE (:P {name:'m'})
CREATE (:P {name:'b'})
MATCH (x:P{name:'m'}),(y:P{name:'a'}) CREATE (x)-[:R]->(y)   // m -> a (physical)
MATCH (x:P{name:'m'}),(y:P{name:'b'}) CREATE (x)-[:R]->(y)   // m -> b (physical)

MATCH (a:P{name:'a'})<-[:R]-(m:P)-[:R]->(b:P{name:'b'}) RETURN m.name
```
Expected: `["m"]` (m -> a and m -> b both hold, so the fork pattern with first-hop-incoming should match). Actual (predicted from reading the code, not yet captured as a passing/failing test): traversal walks the first hop as if `(a)-[:R]->(m)`, i.e. looks for an edge a -> m, which does not exist — likely returns 0 rows or, if `a` happens to coincide with another forward-reachable node, a wrong set.

## Scope note
[...]

Issue #488, opened 2026-08-14
execute_n_hop ignores direction on fixed (non-variable-length) hops — silent wrong rows for 3+-hop chains with a <- hop


ryaker:
## Found while investigating #486 (edge direction inverted)

`execute_n_hop` (crates/sparrowdb-execution/src/engine/hop.rs, `fn execute_n_hop` starting around line 1760) handles chains of 3+ relationships. For **variable-length** hops it explicitly rejects `Incoming`/non-Outgoing direction with `Error::Unimplemented` (lines ~1900-1906: *"`execute_variable_hops` only walks outgoing edges. Reject rather than quietly answering an `<-[:R*1..2]-` pattern with the wrong direction."*) — that's the correct, safe, fail-closed behavior.

But that guard only covers hops where `varlen_per_hop[i].is_some()`. For a **fixed** (non-variable-length) hop within the same chain — the `None` branch at line ~2016-2048 — there is no direction check at all. It always calls `csr_neighbors_labeled(cur_slot, cur_label_id, hop_rel_ids)` (forward-only) and scans `delta_all` for `r.src == (cur_label_id, cur_slot)` (also forward-only), regardless of `pat.rels[hop_idx].dir`.

So a 3+-hop chain with a `<-` on any fixed hop is silently treated as if that hop were `->` — same failure class as #486, in the n-hop function.

## Reproduction sketch

A chain like:
```
MATCH (a:P)-[:R]->(m:P)<-[:R]-(x:P)-[:R]->(b:P) RETURN b.name
```
(4 nodes, 3 fixed hops, second hop incoming) would silently traverse the second hop forward instead of backward, inventing/hiding results the same way #486's single-hop case did.

## Scope note
Not fixed as part of #486 / PR closing it — that PR scoped explicitly to the single fixed-hop case (`execute_one_hop`). `execute_n_hop`'s frontier-based traversal would need per-hop backward-CSR + backward-delta lookup (mirroring what `execute_one_hop`'s fix and the varlen-hop `execute_variable_hops` guard already establish as the safe pattern), or at minimum the same fail-closed `Unimplemented` guard the varlen branch already has, extended to cover fixed hops with non-`Outgoing` direction, until real backward support is added.

## Where to look
- `crates/sparrowdb-execution/src/engine/hop.rs`, `execute_n_hop`, the `match varlen_per_hop[hop_idx] { ... None => { ... } }` branch (fixed-hop neighbour gathering).
[...]

PR #489, opened 2026-08-14
fix(execution): honor EdgeDir::Incoming in execute_one_hop (closes #486)

### LINKED ISSUE #486: CRITICAL: edge direction is inverted — <- returns outbound neighbours, inbound traversal is impossible
Left-pointing relationship patterns do not traverse inbound. Traversal always runs outbound from the anchor regardless of the arrow, so **you cannot ask "what points at this node."**

## Reproduction on `main` (0.1.23)

```
CREATE (:A {id: 'a1'})
CREATE (:B {id: 'b1'})
MATCH (a:A {id:'a1'}), (b:B {id:'b1'}) CREATE (a)-[:R]->(b)
```

| query | expected | actual | |
|---|---|---|---|
| `MATCH (a:A)-[r:R]->(x) RETURN x.id` | `["b1"]` | `["b1"]` | ✅ |
| `MATCH (a:A)<-[r:R]-(x) RETURN x.id` | `[]` | **`["b1"]`** | ❌ |
| `MATCH (b:B)<-[r:R]-(x) RETURN x.id` | `["a1"]` | **`[]`** | ❌ |

Wrong in both directions, which is worse than being ignored:

- `<-` from `A` **invents a relationship that does not exist** — nothing points at `a1`, but the query returns `b1`. A caller asking "what depends on this?" gets a confident wrong answer rather than an empty one.
- `<-` from `B` **hides the relationship that does exist** — `a1` genuinely points at `b1`, and the query returns nothing.

So the arrow is not merely ignored; it is applied as if it always pointed outbound from the anchor.

## Why this is critical

Directed edges exist to answer inbound questions — *what depends on this*, *what supersedes this*, *what references this*, *who reports to this person*. None of those are expressible today, and every one of them currently returns a plausible-looking wrong result rather than an error.

For a knowledge-graph consumer this is close to a total loss of the direction semantics while appearing to work: `->` queries behave correctly, so a test suite that only exercises outbound traversal is green.

## Related, likely the same root cause

`CsrBackward` exists in the storage layer, so the reverse adjacency structure is there — the question is whether the planner ever selects it for a left-pointing pattern, or whether pattern direction is discarded before it reaches traversal. Worth checking whether `MATCH (x)-[r:R]->(b:B)` (anchoring on the right, arrow pointing right) also fails, since that is the same inbound question written the other way and would isolate whether the bug is in direction handling or in anchor selection.

Reported by the Zora integration team, reproduced independently against 0.1.23.

ryaker:
## Summary

- `execute_one_hop` (crates/sparrowdb-execution/src/engine/hop.rs) always bound `src_node_pat = pat.nodes[0]` / `dst_node_pat = pat.nodes[1]` and walked the forward CSR/delta regardless of `RelPattern::dir`. A left-pointing pattern (`(a)<-[:R]-(x)`) was silently executed identically to `(a)-[:R]->(x)`: it invented relationships that don't exist and hid the ones that do, with no error in either case.
- A comment directly above the extraction claimed the incoming case was handled "below" via a swap; no such swap existed anywhere in the function — `EdgeDir::Incoming` was never checked.
- Fix: swap which pattern node plays the physical-source role *before* label resolution, rel-table filtering, and CSR/delta lookup happen. Every later reference in the function goes through `src_node_pat`/`dst_node_pat`, never `pat.nodes[..]` directly, so this one swap is sufficient — it also naturally binds the correct output variable name to each side. This mirrors the swap `execute_one_hop_chunked` (pipeline_exec.rs) already performs correctly for the narrower chunked-pipeline case (both endpoints single-labelled).
- `EdgeDir::Both` (undirected) is untouched — it doesn't take the `Incoming` branch and continues to use its existing separate backward pass (`CsrBackward`) alongside the forward pass.

Closes #486.

## Root cause / anchor-vs-direction

- **Root cause**: `crates/sparrowdb-execution/src/engine/hop.rs:25-26` (pre-fix) — direction was parsed correctly by the parser/AST (`RelPattern::dir` is populated) and even consulted correctly elsewhere (e.g. `execute_two_hop`'s *second* hop, `execute_one_hop_chunked`'s single-labelled case) — but `execute_one_hop`, the fallback "row engine" path used whenever either pattern node is unlabeled (which is exactly how the issue's repro is written), ignored it entirely at traversal time. This matches the task's hypothesis 2: direction was parsed and then ignored at traversal, not lost in the parser/binder.
- **Anchor vs. direction**: `MATCH (x)-[r:R]->(b:B) RETURN x.id` (arrow right, labelled node on the right, anchor unlabeled) already returns the correct `["a1"]` pre-fix — covered by the new `issue_486_anchor_vs_direction_isolation` test, which passes both before and after the fix. This confirms the bug is specifically in `<-` token handling, not in which side is labelled/anchored.

## Test plan

New file: `crates/sparrowdb/tests/regression_486_edge_direction.rs` — 6 e2e tests against a real `GraphDb` on `tempfile::TempDir`, all expected values derived by hand from each fixture (never captured from program output):

- `issue_486_reproduction_table` — the exact 3-row table from the issue.
- `issue_486_anchor_vs_direction_isolation` — the anchor-vs-direction check above.
- `issue_486_multiple_inbound_sources` — two sources pointing at one target; proves *all* inbound edges return, not just one.
- `issue_486_node_with_both_directions` — a node with both an inbound and an outbound edge; `->` and `<-` from it return different, correct sets (catches "direction ignored" as distinct from "direction inverted").
- `issue_486_self_loop_both_directions_coincide` — `(a)-[:R]->(a)`; both directions correctly coincide.
- `issue_486_multiple_rel_types_no_cross_contamination` — `<-[:R]-` does not pick up a `:S` edge to the same node.

Falsifiability, checked against the pre-fix parent commit (`a68ec69`, current `origin/main`) by stashing just the `hop.rs` change:

```
test issue_486_self_loop_both_directions_coincide ... ok
test issue_486_anchor_vs_direction_isolation ... ok
test issue_486_reproduction_table ... FAILED
test issue_486_multiple_inbound_sources ... FAILED
test issue_486_multiple_rel_types_no_cross_contamination ... FAILED
test issue_486_node_with_both_directions ... FAILED

test result: FAILED. 2 passed; 4 failed
```
4/6 fail pre-fix as expected (with the exact `left`/`right` mismatches predicted by the root cause); the 2 that pass pre-fix are the ones designed to isolate anchor-selection (unaffected) and the self-loop degenerate case (both directions coincide even under the bug), not to catch the regression itself.

All 6 pass post-fix.

### Blast radius run
- `cargo fmt --all` — clean.
- `cargo clippy -p sparrowdb -p sparrowdb-execution -p sparrowdb-cypher --all-targets --keep-going` — clean, 0 warnings.
- `uc1_social_graph` (2), `spa_237_unwind_match` (6), `spa_415_unwind_match_mutate` (16 passed / 1 pre-existing unrelated ignore), `regression_slot_identity_root_cause` (9), `reverse_arrow_294` (2), `spa_201_csr_backward` (5), `spa_244_mcp_errors` (4), `spa_299_phase4_parity` (17) — **68/68 passing**, no behavior changes observed.
- Full `cargo test -p sparrowdb --no-fail-fast` (isolated `CARGO_TARGET_DIR=/Volumes/Dev/target-486`, no sibling worktree contamination) run separately — see PR comment / follow-up for final tally, still running at time of PR open given the suite's documented ~60+ min runtime.

## Scope note — related bugs found, filed separately, not fixed here

Same defect class exists in two other traversal functions, neither touched by this PR:

- #487 — `execute_two_hop` never reads `rels[0].dir` (the *first* hop); only the second hop's direction is handled (pre-existing SPA-201/#294 fix). A first-hop `<-` is silently treated as `->`.
- #488 — `execute_n_hop` (3+ hop chains) fail-closes correctly for a variable-length hop with non-Outgoing direction, but has no such guard for a *fixed* hop with non-Outgoing direction — silently treated as outbound.

`execute_variable_length` (the single-hop `*M..N` case) already fail-closes on non-Outgoing direction with `Error::Unimplemented` — confirmed safe, no fix needed.

## Could not verify

- The full unscoped `cargo test -p sparrowdb --no-fail-fast` run had not finished by PR-open time (repo docs put a full run at 60+ minutes); the 68 targeted at-risk tests above are the verified signal for this PR. Will report the full tally as a follow-up comment once it completes.
- #487 and #488's reproduction steps are derived from reading the code, not from a captured failing test — I did not write/run tests confirming those failure modes; they're filed as investigation leads, not confirmed-and-reproduced bugs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected traversal for incoming relationship patterns so left-pointing edges now return the expected connected nodes.
  - Improved handling of mixed edge directions, anchor positions, self-loops, and relationship-type filters.

- **Tests**
  - Added regression coverage for inbound and outbound traversal scenarios, including multiple sources and nodes with edges in both directions.

### DIFF
--- a/crates/sparrowdb-execution/src/engine/hop.rs
+++ b/crates/sparrowdb-execution/src/engine/hop.rs
@@ -22,17 +22,24 @@ impl Engine {
         }
 
         let pat = &m.pattern[0];
-        let src_node_pat = &pat.nodes[0];
-        let dst_node_pat = &pat.nodes[1];
         let rel_pat = &pat.rels[0];
-
         let dir = &rel_pat.dir;
-        // Incoming-only: swap the logical src/dst and recurse as Outgoing by
-        // swapping pattern roles.  We handle it by falling through with the
-        // node patterns in swapped order below.
-        // Both (undirected): handled by running forward + backward passes.
-        // Unknown directions remain unimplemented.
         use sparrowdb_cypher::ast::EdgeDir;
+
+        // Incoming (`(a)<-[:R]-(x)`): the physical edge runs x -> a, so swap
+        // which pattern node plays the "src" (traversal-forward) role before
+        // any label resolution, table filtering, or CSR lookups happen below.
+        // Every reference downstream goes through src_node_pat/dst_node_pat
+        // (never pat.nodes[..] directly), so this single swap is sufficient —
+        // it also naturally binds the correct variable name to each side when
+        // rows are projected at the end of the function.
+        // Both (undirected): handled by running forward + backward passes,
+        // so it keeps the unswapped (nodes[0], nodes[1]) order here.
+        let (src_node_pat, dst_node_pat) = if *dir == EdgeDir::Incoming {
+            (&pat.nodes[1], &pat.nodes[0])
+        } else {
+            (&pat.nodes[0], &pat.nodes[1])
+        };
 
         let src_label = src_node_pat.labels.first().cloned().unwrap_or_default();
         let dst_label = dst_node_pat.labels.first().cloned().unwrap_or_default();
--- a/crates/sparrowdb-execution/src/engine/hop.rs
+++ b/crates/sparrowdb-execution/src/engine/hop.rs
@@ -940,14 +940,28 @@ impl Engine {
         // SPA-263: per-hop CSRs and delta adjacency maps.
         let rel1 = &pat.rels[0];
         let rel2 = &pat.rels[1];
+        // #487: the first hop's direction was never read anywhere in this
+        // function — only the second hop's (`second_hop_incoming` below, the
+        // SPA-201/#294 fix) was. A left-pointing first hop (`(a)<-[:R]-(m)`)
+        // was silently treated as if it had been written `(a)-[:R]->(m)`,
+        // the same failure mode #486 fixed in `execute_one_hop`.
+        let first_hop_incoming = rel1.dir == sparrowdb_cypher::ast::EdgeDir::Incoming;
         let all_rel_tables_2hop = self.snapshot.catalog.list_rel_tables_with_ids();
         let hop1_rel_ids: Vec<u64> = all_rel_tables_2hop
             .iter()
             .filter(|(_, sid, did, rt)| {
                 let type_ok = rel1.rel_type.is_empty() || rt == &rel1.rel_type;
-                let src_ok = *sid as u32 == src_label_id;
-                let dst_ok = *did as u32 == mid_label_id;
-                type_ok && src_ok && dst_ok
+                if first_hop_incoming {
+                    // (src)<-[:R]-(mid): the physical edge runs mid -> src,
+                    // so the catalog table is keyed (src=mid, dst=src).
+                    let src_ok = *sid as u32 == mid_label_id;
+                    let dst_ok = *did as u32 == src_label_id;
+                    type_ok && src_ok && dst_ok
+                } else {
+                    let src_ok = *sid as u32 == src_label_id;
+                    let dst_ok = *did as u32 == mid_label_id;
+                    type_ok && src_ok && dst_ok
+                }
             })
             .map(|(id, _, _, _)| *id)
             .collect();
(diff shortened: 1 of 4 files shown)


ryaker:
## Extended to close #487 and #488 as well

Per team lead's direction: shipping 0.1.24 with `<-` working at one hop and silently empty at two would be more confusing than the current uniform breakage, so this PR now closes #486, #487, and #488 together.

### #487 — `execute_two_hop` ignored the first hop's direction

Only `rels[1].dir` (second hop) was ever read, via the pre-existing SPA-201/#294 `second_hop_incoming` machinery. The first hop was unconditionally treated as Outgoing. Fix adds `first_hop_incoming`, mirrors the existing `merged_bwd_csr`/`delta_adj_bwd` construction for hop1 (`hop1_bwd_csr`/`delta_adj_hop1_bwd`), and introduces a `hop1_mid_candidates(src_slot)` closure used by both the "second hop Incoming" and "second hop Outgoing" branches — so first-hop and second-hop direction are now resolved independently. All 4 combinations (O,O)/(O,I)/(I,O)/(I,I) work; (O,O) and (O,I) are the pre-existing paths, unchanged; (I,O) and (I,I) are new.

### #488 — `execute_n_hop` ignored direction on fixed (non-variable-length) hops

The variable-length branch already fail-closed correctly (`Error::Unimplemented`) on non-Outgoing direction. The **fixed**-hop branch had no such guard and always walked forward. Did the full fix rather than extending the guard: added `Engine::csr_predecessors_labeled` (the backward mirror of the existing `csr_neighbors_labeled`) in `mod.rs`, and branch the fixed-hop neighbour-gathering (CSR lookup + delta-log scan) on each hop's own `dir`. Also fixed `rel_ids_per_hop`'s label-narrowing table lookup, which resolved the catalog table by *pattern* position (src, dst) instead of *physical* direction — silently wrong whenever an Incoming hop's two endpoint labels differ and 2+ tables share the rel type (#429-style ambiguity).

### Confirmed against the team's exact repro before touching code

Fixture `a1 -[:R]-> b1 -[:S]-> c1 -[:T]-> d1`:
- `(c:C)<-[:S]-(b:B)<-[:R]-(a:A)` → was `[]`, now `["a1"]`.
- `(dd:D)<-[:T]-(c:C)<-[:S]-(b:B)<-[:R]-(a:A)` → was `[]`, now `["a1"]`.

Both returned **empty**, not wrong-but-nonempty — worse than #486's failure mode for exactly the reason flagged: for "what transitively depends on this", empty reads as "nothing does".

### Tests

New file: `crates/sparrowdb/tests/regression_487_488_edge_direction.rs` — 10 e2e tests, values derived by hand:
- Team's exact two-hop and three-hop repros, **each run both pre- and post-`CHECKPOINT`** — the delta-log code path and the new persisted-CSR path (`csr_predecessors_labeled`, `hop1_bwd_csr`) are genuinely different code, and no existing MATCH e2e test in this repo called CHECKPOINT before this, so the CSR-backed backward path was previously completely untested.
- Two mixed-direction cases per the request: `(a)<-[:R]-(b)-[:S]->(c)` (new — this is exactly the case a whole-query swap instead of a per-hop one would still fail) and `(a)-[:R]->(b)<-[:S]-(c)` (regression guard on the untouched SPA-201 shape).
- A mixed-direction 3-hop chain for `execute_n_hop` (Out, In, Out).
- Multiple predecessors on a double-Incoming two-hop (proves "every match", not just one).
- All-intermediate-binding check on the 3-hop chain (not just the endpoint).

**Falsifiability**: all 8 of the new tests that exercise the fix (the other 2 are sanity/regression-guard, correct either way) confirmed to return `[]` against this branch's pre-#487/488 parent commit (810143c, i.e. #486's fix alone) by stashing the `hop.rs`/`mod.rs` diff and rerunning — exact output quoted in the commit message and available on request.

### Blast radius

Reran all previously-verified suites (uc1_social_graph, spa_237_unwind_match, spa_415_unwind_match_mutate, regression_slot_identity_root_cause, reverse_arrow_294, spa_201_csr_backward, spa_244_mcp_errors, spa_299_phase4_parity, spa_252_three_hop_binding) plus both new regression files: **84/84 passing**, 1 pre-existing unrelated ignore, no behavior changes. fmt clean, clippy clean (0 warnings) on sparrowdb/sparrowdb-execution/sparrowdb-cypher --all-targets.

Full `cargo test -p sparrowdb --no-fail-fast` is running now in a freshly isolated, never-shared `CARGO_TARGET_DIR` (learned that lesson the hard way earlier in this investigation — will follow up with the tally).

### Could not verify
- Full unscoped suite result pending at comment time (repo docs cite 60+ min for a full run).
- `EdgeDir::Both` (undirected) inside `execute_two_hop` and `execute_n_hop` was already forward-only before this change and remains so — untouched, not regressed, but also not fixed; it's a separate, unfiled gap in the same defect family if anyone hits it (only `execute_one_hop`'s `Both` handling is correct, per #486).

PR #490, opened 2026-08-14
fix(count): COUNT(n)/COUNT(*) now excludes deleted nodes

### LINKED ISSUE #485: count(n) ignores deletions — returns a stale count that survives checkpoint and reopen
`count(n)` does not reflect deleted nodes. It keeps returning the pre-delete total, permanently.

## Reproduction on `main` (0.1.23)

```
CREATE (:U {id: 'x'})
CREATE (:U {id: 'y'})

MATCH (n:U) RETURN count(n)      -> 2      correct
MATCH (n:U {id:'x'}) DELETE n
MATCH (n:U) RETURN count(n)      -> 2      WRONG, expected 1
MATCH (n:U) RETURN n.id          -> ["y"]  correct — 'x' really is gone

MATCH (n:U) DELETE n                        delete the rest
MATCH (n:U) RETURN count(n)      -> 2      WRONG, expected 0
MATCH (n:U) RETURN n.id          -> []     correct — all gone

checkpoint()
MATCH (n:U) RETURN count(n)      -> 2      still wrong
-- drop the handle, reopen the database --
MATCH (n:U) RETURN count(n)      -> 2      WRONG
MATCH (n:U) RETURN n.id          -> []     correct
```

Row projection is right at every step, including after a reopen — so the deletions are real and durable. `count(n)` alone is wrong, and it stays wrong across checkpoint and restart, which suggests it reads a slot high-water mark or a catalog row-count rather than counting live nodes.

## Why this is more dangerous than an ordinary wrong answer

It is silently wrong in the direction that makes callers believe data still exists. An external integrator hit exactly this and concluded **"DELETE is a no-op — there is no way to remove a node from the graph, unbounded growth, no retention path, and right-to-erasure is unimplementable."** That conclusion is false — deletion works — but it is entirely reasonable given `count()` is the natural way to check, and it was serious enough to be escalated as a compliance blocker.

A count that over-reports also breaks the obvious pagination, quota, and "is this empty" checks, and it is the metric most likely to be wired into a dashboard or an alert.

## Scope not yet established

- Whether `count(*)` behaves the same as `count(n)`
- Whether other aggregates (`collect`, `sum`, `avg`) share the source
- Whether edge counts have the equivalent problem after `DETACH DELETE`
- Whether the stale value comes from the catalog's label row-count or the node store's high-water mark — `build_label_row_counts_from_disk` is the first place to look

Reported by the Zora integration team; deletion semantics re-verified independently against 0.1.23 before filing.

ryaker:
## Summary
- Fixes #485: `COUNT(n)` / `COUNT(*)` over a labeled pattern permanently overcounted once any node of that label was deleted, and stayed wrong across `checkpoint()` and reopen.
- Root cause: `label_row_counts` (the cache backing SPA-197's O(1) COUNT-label fast-path) was seeded from `NodeStore::hwm_for_label`, a high-water mark of slots ever *allocated*. `delete_node`/`detach_delete_node` tombstone the slot in place (write `u64::MAX` into `col_0`) instead of freeing it, so the HWM never shrinks — and `build_label_row_counts_from_disk` (which populates `label_row_counts`) is also the function `GraphDb::open()` calls to seed the cache, so the staleness survived process restart.
- Fix: added `NodeStore::live_count_for_label`, which subtracts tombstoned `col_0` slots from the HWM — the same visibility rule ordinary scans already apply via `is_node_tombstoned`. `build_label_row_counts_from_disk` now calls it instead of `hwm_for_label` directly.

## Blast radius (verified by hand, not assumed)
| Aggregate | Pre-fix | Notes |
|---|---|---|
| `COUNT(n)` (labeled) | **broken** | fixed here |
| `COUNT(*)` (labeled) | **broken** | same fast-path, same cache; fixed here |
| `COUNT(n)` no label / with WHERE | fine | falls through to normal scan (tombstone-aware) |
| `collect`, `sum`, `avg`, `min`, `max` | fine | never hit the label-count fast-path |
| edge count after `DETACH DELETE` | fine | separate delta-log/CSR code path |
| delete-then-create (slot reuse) | n/a | `create_node` is append-only, never reuses a tombstoned slot — covered by a new test anyway |

`GraphDb::db_counts()` (backs the `sparrowdb info` CLI) has the *same* HWM-based staleness in a separate code path (`db.rs::db_counts`), and has its own test (`spa_217_info_counts.rs::db_counts_hwm_includes_deleted_nodes`) explicitly documenting the current stale behavior as intended. Left untouched — out of scope for #485 as filed (which is specifically about Cypher `COUNT(n)`). Worth a follow-up issue if `sparrowdb info` should also reflect deletions.

## Test plan
New file: `crates/sparrowdb/tests/regression_485_stale_count.rs` — all values derived by hand from the CREATE/DELETE statements, not captured from program output.
- [x] count after deleting one of several nodes
- [x] count after deleting all nodes of a label (must be exactly 0)
- [x] count survives `checkpoint()` + drop handle + reopen (the exact repro from the issue)
- [x] count after delete-then-create (HWM growth past a tombstoned slot)
- [x] `COUNT(*)` after delete (same fast-path, separate branch)
- [x] a control label untouched by any delete is unaffected

Confirmed every new test **fails against the pre-fix parent commit** (`a68ec69`) with the exact stale-count symptom, by stashing the source changes and re-running.

Also ran (all green): `spa_197_count_label_fastpath`, `spa_194_count_node_var`, `spa_242_count_rel_var`, `spa_172_count_distinct`, `spa_272_q7_count_fastpath`, `spa_272_degree_cache`, `spa_217_info_counts`, `spa_aggregation`, `phase7_mutation`, `spa_187_column_slot_alignment`.

`cargo fmt` and `cargo clippy -p sparrowdb -p sparrowdb-execution -p sparrowdb-storage --all-targets --keep-going` are clean.

Full `cargo test -p sparrowdb -p sparrowdb-execution -p sparrowdb-storage --no-fail-fast` was kicked off (the suite runs 60+ min per repo notes); will report results in a follow-up comment.

### DIFF
--- a/crates/sparrowdb-storage/src/node_store.rs
+++ b/crates/sparrowdb-storage/src/node_store.rs
@@ -73,6 +73,12 @@ const TAG_FLOAT: u8 = 0x03;
 /// Maximum bytes that fit inline in the 7-byte payload (one byte is the tag).
 const MAX_INLINE_BYTES: usize = 7;
 
+/// Deletion tombstone sentinel written into `col_0` by `tombstone_node`.
+/// Mirrors the private `TOMBSTONE` constant in `property_index.rs` /
+/// `text_index.rs` — kept separate because those modules read the value
+/// back out of `NodeStore` rather than defining it themselves.
+const TOMBSTONE_SENTINEL: u64 = u64::MAX;
+
 impl Value {
     /// Encode as a packed `u64` for column storage.
     ///
@@ -641,11 +647,39 @@ impl NodeStore {
     /// Return the high-water mark (slot count) for a label.
     ///
     /// Returns `0` if no nodes have been created for that label yet.
+    ///
+    /// This is the number of slots ever *allocated* for the label — it is
+    /// never decremented by [`tombstone_node`](Self::tombstone_node), so it
+    /// overcounts once any node of this label has been deleted. Use
+    /// [`live_count_for_label`](Self::live_count_for_label) when the answer
+    /// needs to reflect deletions (e.g. `COUNT(n)`).
     pub fn hwm_for_label(&self, label_id: u32) -> Result<u64> {
         if let Some(&h) = self.hwm.get(&label_id) {
             return Ok(h);
         }
         self.load_hwm(label_id)
+    }
+
+    /// Return the number of *live* (non-deleted) nodes for a label (#485).
+    ///
+    /// `hwm_for_label` counts every slot ever allocated; deleted nodes are
+    /// tombstoned in place (`col_0` slot set to `u64::MAX`) rather than
+    /// freeing the slot, so the HWM alone overcounts once deletions have
+    /// happened. This reads `col_0` once and subtracts the tombstoned slots
+    /// from the HWM, matching the row-visibility rule `is_node_tombstoned`
+    /// already applies to ordinary scans.
+    pub fn live_count_for_label(&self, label_id: u32) -> Result<u64> {
+        let hwm = self.hwm_for_label(label_id)?;
+        if hwm == 0 {
+            return Ok(0);
+        }
+        let col0 = self.read_col_all(label_id, 0)?;
+        let tombstones = col0
+            .iter()
+            .take(hwm as usize)
+            .filter(|&&v| v == TOMBSTONE_SENTINEL)
+            .count() as u64;
+        Ok(hwm.saturating_sub(tombstones))
     }
 
     /// Discover all column IDs that currently exist on disk for `label_id`.
@@ -1168,7 +1202,8 @@ impl NodeStore {
         // Seek to the slot and write the tombstone value.
         f.seek(SeekFrom::Start(slot as u64 * 8))
             .map_err(Error::Io)?;
-        f.write_all(&u64::MAX.to_le_bytes()).map_err(Error::Io)
+        f.write_all(&TOMBSTONE_SENTINEL.to_le_bytes())
+            .map_err(Error::Io)
     }
 
     /// Overwrite the value of a single column for an existing node.
--- a/crates/sparrowdb/src/db.rs
+++ b/crates/sparrowdb/src/db.rs
@@ -3075,28 +3075,37 @@ impl GraphDb {
         }
     }
 
-    /// Return `(node_count, edge_count)` by summing the high-water marks
+    /// Return `(node_count, edge_count)` by summing the live node count
     /// across all catalog labels (nodes) and delta-log record counts across
     /// all registered relationship tables (edges).
     ///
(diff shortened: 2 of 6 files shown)


ryaker:
## Extended scope: also closes #491

Per team lead's call for this release ("fix all"), extended this PR to cover `GraphDb::db_counts()` (backs the `sparrowdb info` CLI), which had the identical HWM-based staleness in a separate code path — filed as #491.

**Fix**: `db_counts()` now calls `NodeStore::live_count_for_label` (already added for #485) instead of summing raw `hwm_for_label`.

**Test fix**: `spa_217_info_counts.rs::db_counts_hwm_includes_deleted_nodes` asserted the bug as intended behavior, and its only assertion was inside `if delete_result.is_ok()` — so it could never fail even if DELETE silently broke. Rewrote as `db_counts_excludes_deleted_nodes`: DELETE is now asserted to succeed (`.expect(...)`, no guard), and the count is asserted to drop to 0 unconditionally. Also corrected the doc comment on `db_counts()` — there is no compaction/GC path that ever reconciled the old value; it stayed wrong indefinitely, including across checkpoint/reopen.

**New tests** (`regression_491_stale_db_counts.rs`, 4 tests): partial delete, delete-all, checkpoint+reopen (the #485-equivalent repro), and a control label unaffected by a delete on a sibling label. All 5 new/changed assertions (4 new + 1 rewritten) confirmed to fail against the pre-fix db.rs with the exact stale-count symptom, e.g.:
```
left: 2
right: 0
```
(2 = stale pre-delete HWM, 0 = hand-derived expected after delete-all)

**Closed could-not-verify items from the original report**:
- Language bindings: `sparrowdb-node`'s `execute()` calls straight through to `GraphDb::execute()` — no COUNT-specific code in the napi layer. `cargo check -p sparrowdb-node` passes clean against the fix.
- External dependency on the stale value: `grep -rn "db_counts\|node_count\|edge_count"` across `sparrowdb-mcp`, the `sparrowdb-mcp` binary, and `sparrowdb-cli` shows the MCP servers hardcode `node_count`/`edge_count` to `null` ("will be populated once the catalog API is promoted") — they never call `db_counts()`. Only the CLI's `info` command does, which is exactly the surface #491 was filed against. No hidden dependency found.

**Full-suite result** (`cargo test -p sparrowdb -p sparrowdb-execution -p sparrowdb-storage --no-fail-fast`, run against the #485 fix before the #491 commit): **1411 passed, 0 failed** across 187 test binaries — clean run, no regressions.

`cargo fmt` and `cargo clippy -p sparrowdb -p sparrowdb-execution -p sparrowdb-storage --all-targets --keep-going` clean. `cargo check -p sparrowdb` passes. Still not merged.

PR #499, opened 2026-08-15
Fix 2-hop MATCH with unlabeled terminal node

### LINKED ISSUE #493: 2-hop MATCH with an unlabeled terminal node fails with the uninformative error "not found"
A two-hop `MATCH` whose final node has no label errors out, and the error text is the bare string `not found` — no clause, no token, no position, no mention of what wasn't found.

## Reproduction on `main` (0.1.23) — direction-agnostic

Fixture: `a1 -[:R]-> b1`, `a1 -[:S]-> s1`

| query | result |
|---|---|
| `MATCH (s:S)<-[:S]-(a:A)-[:R]->(x) RETURN x.id` | **`Err("not found")`** |
| `MATCH (s:S)<-[:S]-(a:A)<-[:R]-(x) RETURN x.id` | **`Err("not found")`** |
| `MATCH (s:S)<-[:S]-(a:A)<-[:R]-(x:B) RETURN x.id` | `[]` — works once labelled |
| `MATCH (b:B)<-[:R]-(x) RETURN x.id` | `["a1"]` — one hop with an unlabeled tail is fine |

So it is specifically **two hops + unlabeled terminal**, and it affects outbound and inbound identically. Confirmed present on `main` before the #486/#487/#488 direction work, so it is pre-existing and unrelated to those.

Single-hop unlabeled works, which makes this an inconsistency rather than a documented restriction — a user who successfully writes `(b:B)<-[:R]-(x)` has no reason to expect the two-hop form to fail.

## The error message is the worse half

`not found` gives a caller nothing to act on. Compare the genuinely good diagnostics elsewhere in this codebase:

> `variable-length paths require a relationship type: use [:R*] not [*]`

> `HNSW index generation conflict: … is at generation 2 but this handle loaded generation 0. Another writer has replaced the index since it was opened; saving would discard their vectors. Reopen the database…`

An external integrator hit the same bare `not found` on `RETURN DISTINCT type(r)` and reported it as uninformative (#480, S3-1). Two unrelated features surfacing the identical contentless string suggests a shared catch-all worth auditing — `grep` for where this error is constructed and check what else routes through it.

## Suggested fix

Two parts, and the second matters even if the first is deferred:
1. Support the unlabeled terminal in two-hop patterns, matching one-hop behaviour.
2. If it must stay unsupported, say so: name the clause and the reason, the way the varlen error does. A restriction a user can read is a limitation; a bare `not found` is a bug report.

Found while verifying the direction fixes in #489.

ShlokShar:
Fixes #493 by routing 2-hop patterns with an unlabeled terminal node through the generic N-hop executor, which resolves the final hop's label from stored edges instead of requiring a catalog label lookup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed two-step relationship queries involving unlabeled terminal nodes, preventing valid results from being incorrectly reported as not found.
  * Improved incoming and outgoing traversal accuracy, including relationship direction, labels, and metadata.
  * Improved query results for property projections, aggregations, relationship filters, and distinct relationship counts.
  * Preserved existing behavior for property-only projections.

### DIFF
--- a/crates/sparrowdb-execution/src/engine/hop.rs
+++ b/crates/sparrowdb-execution/src/engine/hop.rs
@@ -838,6 +838,14 @@ impl Engine {
         // nodes[2] is the fof (friend-of-friend) / anchor-B in Q8
         let fof_node_pat = &pat.nodes[2];
 
+        // If the terminal node is unlabeled, the generic N-hop executor already
+        // recovers the actual label of every final hop from the stored edges.
+        // Route there instead of requiring fof_label_id and returning a bare
+        // "not found".
+        if fof_node_pat.labels.is_empty() {
+            return self.execute_n_hop(m, column_names);
+        }
+
         let src_label = src_node_pat.labels.first().cloned().unwrap_or_default();
         let fof_label = fof_node_pat.labels.first().cloned().unwrap_or_default();
         let src_label_id = self
--- a/crates/sparrowdb-execution/src/engine/hop.rs
+++ b/crates/sparrowdb-execution/src/engine/hop.rs
@@ -2018,6 +2018,9 @@ impl Engine {
         // Reusable neighbor buffer for the variable-length traversal.
         let mut neighbors_buf: HashSet<(u64, u32)> = HashSet::new();
 
+        let use_agg = has_aggregate_in_return(&m.return_clause.items);
+        let use_eval_projection = needs_node_ref_in_return(&m.return_clause.items);
+        let mut raw_rows: Vec<HashMap<String, Value>> = Vec::new();
         let mut rows: Vec<Vec<Value>> = Vec::new();
 
         for src_slot in 0..hwm_src {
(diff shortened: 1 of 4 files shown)


ryaker:
Reviewed and verified independently — thank you, this is a good contribution and it merges.

**Verified it does not regress the direction work.** This replaces `Engine::csr_predecessors_labeled`, added only hours ago in #489, with a per-table backward-CSR lookup, so that was the first thing I checked. I ran my own direction matrix rather than relying on the suite, pre- and post-CHECKPOINT since those are genuinely different code paths:

| | pre-ckpt | post-ckpt |
|---|---|---|
| 1-hop out / in / negative | ✅ ✅ ✅ | ✅ ✅ ✅ |
| 2-hop all-inbound | ✅ | ✅ |
| 3-hop all-inbound | ✅ | ✅ |

All 23 tests across `regression_486_edge_direction`, `regression_487_488_edge_direction` and your new `regression_499_unlabeled_terminal` pass. Your edits to the existing regression file are comment and assertion-*message* only — no assertion logic touched, which is exactly right.

**The fix works for the case it names:**

```
MATCH (a:A)-[:R]->(b:B)-[:S]->(x) RETURN x.id
  before: ERR "not found"
  after:  ["c1"]
```

**Scope note, not a criticism.** #493 covers three shapes and this closes one of them. The other two are unchanged:

| case | before | after |
|---|---|---|
| unlabeled **tail** — this PR | `ERR not found` | **`["c1"]`** ✅ |
| unlabeled **head**: `MATCH (x)-[:R]->(b:B)-[:S]->(c:C)` | `ERR not found` | `ERR not found` |
| unlabeled **middle**: `MATCH (a:A)-[:R]->(m)-[:S]->(c:C)` | `[]` | `[]` |

Your PR title says "unlabeled terminal node", so the scope is honestly stated — I'm noting it only because the body says "Fixes #493", which will auto-close the issue on merge. I'll reopen it with the remaining two cases documented rather than edit your description.

The middle case is the one I'd flag as most worth a follow-up: it returns `[]` **silently** rather than erroring, and only when the pattern's node labels differ. A wrong answer with no error is worse than the `not found` you fixed.

**On CI:** `claude-review` shows as failed — that is the fork security boundary, not your code. Its log reads `Secret source: None`, so it had no API token. Fork PRs correctly get no repository secrets here. Everything that can meaningfully run on a fork passes.

Issue #500, opened 2026-08-15
Statement::Pipeline is never checked by is_mutation — CALL{} mutation inside a pipeline can silently no-op like #478


ryaker:
Found while auditing the `Statement` variant list against `is_mutation` for #478.

`Statement::is_mutation` (`crates/sparrowdb-cypher/src/ast.rs`, formerly `Engine::is_mutation` in `crates/sparrowdb-execution/src/engine/mod.rs`) does not have an arm for `Statement::Pipeline` at all — it falls through to `_ => false`.

A `Statement::Pipeline` can contain a `PipelineStage::CallSubquery { subquery, .. }` whose `subquery` is a mutation, e.g.:

```cypher
MATCH (n:Person) WITH n CALL { CREATE (:Audit {x: 1}) } RETURN n
```

Confirmed this parses: `parse_call_subquery_stage` (`crates/sparrowdb-cypher/src/parser.rs:2846`) calls `self.parse_statement()` for the subquery body inside a pipeline `CALL { }` stage with no restriction on which statement kind is allowed there — a bare `CREATE` parses fine.

At execution time this stage is run via `execute_pipeline_call_subquery_stage` (`crates/sparrowdb-execution/src/engine/subquery.rs:177`), which calls `self.execute_read_stmt(subquery)` — the same non-transactional read-engine path that caused #478. Because `is_mutation` never inspects `Statement::Pipeline`, the whole pipeline statement is dispatched through the read-only engine in `GraphDb::execute` regardless of what's inside the `CALL {}` stage.

Expected symptom, by analogy with #478: the statement reports success (or partial success) and the `CREATE` inside the `CALL {}` stage does not durably persist.

Not fixed as part of #478 — that PR fixed the `UNION`-over-mutation case specifically (rejecting at bind time) and promoted the enumeration to a single `Statement::is_mutation` method, but deliberately left `Pipeline` out of scope per plan. This should get the same treatment: either recursively check `PipelineStage::CallSubquery` bodies in `Statement::is_mutation` and route through the write-transaction path, or reject a mutating `CALL {}` stage inside a pipeline at bind time.

cc #478

ryaker:
Closing — **this does not reproduce.** Every pipeline `CALL { }` form containing a mutation already fails closed on `main` (0.1.24), verified by running them:

| query | result on main |
|---|---|
[...]

PR #501, opened 2026-08-15
fix(cypher): reject UNION over a mutating clause (closes #478)

### LINKED ISSUE #478: UNION-wrapped CREATE bypasses the is_mutation gate and durably writes nothing
`Engine::is_mutation` (`crates/sparrowdb-execution/src/engine/mod.rs:1048-1064`) enumerates `Merge`, `MatchMergeRel`, `MatchMutate`, `MatchCreate`, `Create`, and recursive `CallSubquery` — but **not** `Statement::Union`, which falls to `_ => false`.

So a query shaped:

```cypher
CREATE (:A {x: 1}) RETURN 1
UNION
CREATE (:B {y: 2}) RETURN 2
```

skips the mutation gate at `db.rs:555`, goes through the read-engine dispatch, and `execute_union` recurses into `execute_bound(Statement::Create)` → `mutation.rs`'s non-transactional `execute_create`.

Observed: the statement returns successfully and **zero nodes exist on disk afterwards**. The read path's `NodeStore` instance does not durably persist through this route, so the writes are silently discarded. No error is raised.

A `CREATE` that reports success and writes nothing is a silent data-loss shape, even if the current blast radius is small because the route is unusual. The two candidate fixes point in opposite directions and the choice should be deliberate: either reject `UNION` containing a mutation at parse or bind time (Cypher does not define `UNION` over write clauses), or add `Union` to `is_mutation` so it routes to the transactional writer. Rejecting is probably right.

Note this also means `mutation.rs::execute_create`'s documented non-atomicity (the node is written by `create_node()` before the FTS block, with no rollback if that block fails) has no *observable* consequence through this route today — nothing durably persists either way. That should not be relied on as a safety property; it is a second bug masking the first.

Found during adversarial review of #469. Pre-existing on `main`, unrelated to that PR.

ryaker:
## Summary

`Engine::is_mutation` enumerated `Merge`, `MatchMergeRel`, `MatchMutate`, `UnwindMatchMutate`, `MatchCreate`, `Create`, and recursive `CallSubquery` — but not `Statement::Union`, which fell through to `_ => false`. A query shaped `CREATE (...) RETURN 1 UNION CREATE (...) RETURN 2` skipped the mutation gate in `GraphDb::execute`, was dispatched through the read-only engine, and `execute_union` recursed into the non-transactional `mutation.rs::execute_create`. The statement returned `Ok` and **zero nodes existed on disk afterwards** — silent data loss.

## Direction chosen: reject, don't route

Cypher does not define `UNION` over mutating branches. Rather than adding `Union` to the mutation enumeration and routing it through the write-transaction path (a shape nobody asked for, and one that would need real thought about `WriteTx` semantics across two independently-executed branches), this rejects at **bind time**:

- `sparrowdb-cypher::binder::bind`'s `Statement::Union` arm now checks `u.left.is_mutation() || u.right.is_mutation()` and returns `Err(InvalidArgument(...))` before either side executes. Nesting (`UNION` of `UNION`s) is covered for free since `bind()` recurses into each side.
- `Statement::is_mutation` is promoted to a method on `Statement` itself (`sparrowdb-cypher::ast`) — the single source of truth. `Engine::is_mutation` now just delegates to it. This directly addresses the root cause: the enumeration living in two places (or getting extended ad hoc) is exactly the shape that goes stale, which is how `Union` got missed in the first place.

## Neighbourhood audit

Checked the full `Statement` variant list against `is_mutation`. Found one more real gap, filed separately rather than fixed here (out of scope per plan): **#500** — `Statement::Pipeline` isn't checked by `is_mutation` at all, and can carry a `PipelineStage::CallSubquery` whose inner statement is a mutation (`MATCH ... WITH ... CALL { CREATE ... } ... RETURN ...`). Confirmed this parses (`parse_call_subquery_stage` calls `parse_statement()` unrestricted) and executes via the same non-transactional `execute_read_stmt` path as this bug.

## `execute_create`'s non-atomicity

`mutation.rs::execute_create` writes the node via `create_node()` before the later FTS block with no rollback on failure — documented as non-transactional. This fix does **not** make that non-atomicity observable: `UNION`-with-`CREATE` is now rejected before `execute_create` is ever reached via that route, so the non-transactional read-path `CREATE` remains unreachable through `UNION`, not made durable. Not relying on that as a safety property — it's a separate latent issue in `mutation.rs`.

## Testing

New `crates/sparrowdb/tests/regression_478_union_create.rs`, e2e against a real `GraphDb` on `tempfile::TempDir`. Every durability assertion is **after `drop`-ing the handle and reopening** — querying the same handle can pass while nothing reached disk, which is how #478 hid.

- `union_of_two_creates_is_rejected_not_silently_dropped` — UNION of two CREATEs
- `union_mixing_create_and_match_is_rejected` — CREATE mixed with plain MATCH, both orderings
- `union_all_of_two_creates_is_rejected` — UNION ALL
- `read_only_union_still_works` — plain read UNION ALL is unaffected
- `union_with_merge_is_rejected` — MERGE (a different mutation shape) is also caught; note this one already errored pre-fix, just via an unrelated confusing message (see test docstring) — this fix replaces that with the correct UNION-specific explanation

All 4 mutation-rejection tests confirmed to **fail against the pre-fix commit** (`git stash` of the 3 source files, tests run against `origin/main`):
```
union_of_two_creates_is_rejected_not_silently_dropped ... FAILED
  "got Ok: Ok(QueryResult { columns: [], rows: [] })"
union_all_of_two_creates_is_rejected ... FAILED
  "got Ok: Ok(QueryResult { columns: [], rows: [] })"
union_mixing_create_and_match_is_rejected ... FAILED
  "got Ok: Ok(QueryResult { columns: [\"n.name\"], rows: [[String(\"Alice\")]] })"
union_with_merge_is_rejected ... FAILED
  "got: InvalidArgument(\"mutation statements must be executed via execute_mutation\")"
```
All 5 pass post-fix. Also ran the existing UNION, CREATE, MERGE, CallSubquery, batch-write, and parameterized-CREATE suites (~100 tests across `spa_132_union`, `call_subquery`, `match_after_create`, `merge_node`, `security_480_parameterized_create`, `spa_137/168/182/183/211/215/224/233/235_234/243/415`) — all pass, no regressions.

`cargo fmt` clean (no diff), `cargo clippy -p sparrowdb -p sparrowdb-execution -p sparrowdb-cypher --all-targets --keep-going` clean, `cargo check -p sparrowdb` clean.

## Could not verify

- Did not run the full workspace suite (`cargo test --workspace --exclude sparrowdb-ruby`) — the repo's own docs note it can exceed 60 minutes; ran targeted UNION/CREATE/MERGE/subquery suites instead (~100 tests, all green). A full run before merge would give broader confidence but wasn't done here.
- Did not independently verify #500 (the Pipeline gap) is otherwise unreachable in production usage — only confirmed it parses and takes the vulnerable code path; did not check whether any current caller composes pipelines with mutating `CALL {}` stages in practice.

## Test plan
- [x] New regression suite passes post-fix, fails pre-fix (quoted above)
- [x] Existing UNION/CREATE/MERGE/subquery suites pass with no regressions
- [x] `cargo fmt`, `cargo clippy`, `cargo check -p sparrowdb` clean

Closes #478. Filed #500 as a follow-up for the Pipeline gap found during the audit.

### DIFF
--- a/crates/sparrowdb-cypher/src/ast.rs
+++ b/crates/sparrowdb-cypher/src/ast.rs
@@ -594,3 +594,39 @@ pub enum Statement {
         name: String,
     },
 }
+
+impl Statement {
+    /// Returns `true` if executing this statement can durably mutate graph
+    /// state, and therefore must be routed through the transactional
+    /// `WriteTx` path rather than the read-only engine dispatch.
+    ///
+    /// This is the single source of truth for that classification — do not
+    /// duplicate this match elsewhere. A statement shape that can reach a
+    /// mutation but is missing here is a silent-data-loss bug: the caller
+    /// takes the read-only path, executes a `CREATE`/`MERGE`/`SET`/`DELETE`
+    /// against a non-transactional `NodeStore`, and reports success while
+    /// nothing durable happens (issue #478).
+    ///
+    /// `Union` is deliberately *not* listed as a mutation here: Cypher does
+    /// not define `UNION` over mutating branches, so the binder rejects a
+    /// `UNION` with a mutating side at bind time (see `bind()` in
+    /// `binder.rs`) rather than trying to route it through the write path.
+    pub fn is_mutation(&self) -> bool {
+        match self {
+            Statement::Merge(_)
+            | Statement::MatchMergeRel(_)
+            | Statement::MatchMutate(_)
+            | Statement::UnwindMatchMutate(_)
+            | Statement::MatchCreate(_) => true,
+            // All standalone CREATE statements must go through the
+            // write-transaction path to ensure WAL durability and correct
+            // single-writer semantics, regardless of whether edges are present.
+            Statement::Create(_) => true,
+            // Recursively check CALL { } subqueries so that a mutation inside
+            // the braces is routed through the write-transaction path rather than
+            // being silently dispatched via the read path and failing late.
+            Statement::CallSubquery { subquery, .. } => subquery.is_mutation(),
+            _ => false,
+        }
+    }
+}
--- a/crates/sparrowdb-cypher/src/binder.rs
+++ b/crates/sparrowdb-cypher/src/binder.rs
@@ -64,7 +64,25 @@ pub fn bind(stmt: Statement, catalog: &Catalog) -> Result<BoundStatement> {
         Statement::OptionalMatch(_) => {}
         Statement::MatchOptionalMatch(mom) => bind_match_optional_match(mom, catalog)?,
         // UNION: bind both sides independently.
+        //
+        // Cypher does not define UNION over mutating clauses. Reject here
+        // rather than letting a mutating branch (CREATE, MERGE, MATCH...SET,
+        // MATCH...DELETE) reach the read-only engine dispatch: that path runs
+        // against a non-transactional NodeStore, so the statement would
+        // report success while durably writing nothing (issue #478). This
+        // check must run on *both* sides — `UNION` is left/right-symmetric
+        // and a mutation can appear on either branch, or nested arbitrarily
+        // deep in a `UNION`-of-`UNION` chain (each nested `Union` is bound
+        // recursively via this same arm, so nesting is covered for free).
         Statement::Union(u) => {
+            if u.left.is_mutation() || u.right.is_mutation() {
+                return Err(sparrowdb_common::Error::InvalidArgument(
+                    "UNION cannot contain a mutating clause (CREATE, MERGE, \
+                     MATCH ... SET, MATCH ... DELETE); each UNION branch must \
+                     be a read-only query"
+                        .into(),
+                ));
+            }
             bind((*u.left).clone(), catalog)?;
             bind((*u.right).clone(), catalog)?;
         }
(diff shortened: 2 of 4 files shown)


ryaker:
@coderabbitai full review

The check on this PR currently reads `Review rate limited` — no review has run yet.

Two things worth your attention:

1. **The structural change matters more than the fix.** `is_mutation` was duplicated between `sparrowdb-cypher` and `sparrowdb-execution`, and this PR promotes it to `Statement::is_mutation(&self)` as a single source of truth. The bug it fixes (#478 — `UNION` over a `CREATE` silently wrote nothing) existed because those two lists could drift. Please check whether any caller still classifies mutations independently.

2. **A second instance of that drift is already known and NOT fixed here** — #502: `is_mutation` counts `CallSubquery` as a mutation, but the dispatch at `db.rs:556` has no arm for it, so a bare `CALL { CREATE (:X) }` reaches `_ => unreachable!()` and panics. If this PR's single-source-of-truth work makes that pairing compiler-checkable, say so — that would retire the class rather than the instance.

PR #512, opened 2026-08-15
fix(storage): DELETE dangling edges + DETACH DELETE cross-table pollution (#436)

### LINKED ISSUE #436: MATCH (n:Label) DELETE n leaves dangling edges; DETACH DELETE fails with __SO_HAS_PROPERTY not found
Two node-deletion bugs, found while building instance-data RDF export against SparrowDB `21b7f5e`. Both verified **durable across `checkpoint()` and reopen**.

## 1. `DELETE` leaves dangling edges

```cypher
MATCH (n:Person) DELETE n
```

Removes the nodes, but their incident edges survive and are still returned:

```cypher
MATCH (a)-[r:WORKS_FOR]->(b) RETURN id(a), id(b)   -- still yields the deleted endpoints
```

Any consumer traversing edges afterwards asserts facts about entities that no longer exist. The RDF exporter had to grow a defensive guard — it skips edges whose endpoints are gone and counts them in `ExportReport::dangling_edges` — to avoid emitting triples about deleted subjects.

## 2. `DETACH DELETE` fails outright

```
relationship type '__SO_HAS_PROPERTY' not found in catalog for labels (0, N)
```

`DETACH DELETE` appears to attempt deletion across **every** relationship type in the catalog rather than only those incident to the node, so it trips over ontology-internal `__SO_*` relationship types that are not defined for the node's label pair.

The two compound: `DETACH DELETE` is the documented way to avoid dangling edges, and it does not work here.

## Impact

Deleting a node cannot currently be done cleanly through Cypher. `DELETE` corrupts the edge set; `DETACH DELETE` errors. Anything that then reads relationships gets results referring to deleted nodes — silent, and only visible if you cross-check endpoints against node existence.

## Scope

- Reproduced on `21b7f5e` (v0.1.22 plus unreleased fixes), macOS aarch64.
- Both persist across `checkpoint()` and reopen, so this is not a delta-log visibility artifact.
- Not caused by any PR merged today.

## Note

Worth checking whether the tombstone path filters edges at all — a related known gap is that tombstoned nodes are not filtered from traversal (noted while fixing #427).

ryaker:
## Summary

Fixes #436 — two related node-deletion bugs, both rooted in `WriteTx` treating a CSR slot number as unambiguous node identity (the same root-cause class already fixed on the *read* path for #415/#427/#429/#431 — see `regression_slot_identity_root_cause.rs` — now fixed on the *write* path).

**1. `DELETE` (no `DETACH`) left checkpointed edges dangling.**
`WriteTx::delete_node`'s `NodeHasEdges` safety check indexed CSR forward/backward files with the *full packed* `NodeId` (label bits included) instead of the label-relative slot. For any label other than label 0 this value overflows `CsrForward`/`CsrBackward::n_nodes`, so `neighbors()`/`predecessors()` silently returned `&[]` — the check never fired for checkpointed edges. A node with a checkpointed edge could be `DELETE`d, leaving the edge pointing at a tombstoned node.

**2. `DETACH DELETE` failed with `relationship type '...' not found in catalog for labels (...)`.**
`WriteTx::detach_delete_node` correctly stripped the label bits before indexing, but consulted *every* registered relationship table's CSR regardless of whether the deleted node's own label matched that table's declared src/dst label. When the node's slot number coincidentally aligned with a valid slot in an unrelated table, the code attributed that table's adjacency data to the node — including manufacturing a nonexistent self-loop — and the bogus edge tuple then failed to resolve in the catalog, surfacing exactly the reported error.

## Fix

Both methods now only consult a relationship table's forward CSR when the node's label matches that table's declared src label, and its backward CSR when it matches the dst label — mirroring the label-filtered lookup pattern already established on the read path.

## Verification

`crates/sparrowdb/tests/regression_436_dangling_edges.rs` adds 5 e2e tests against a real `GraphDb` on `tempfile::TempDir`, asserting on every `Result` (no `.ok()`), covering both `DELETE` and `DETACH DELETE`, both directions of traversal, pre- and post-`checkpoint()`, and full reopen. All expected values were derived by hand from each fixture's `CREATE` calls.

Confirmed against the pre-fix commit (712ace4): the 3 bug-catching tests fail with the exact reported symptoms —
- `delete_without_detach_errors_post_checkpoint`: DELETE silently succeeds instead of returning `NodeHasEdges`
- `detach_delete_removes_node_and_edge_post_checkpoint` / `detach_delete_does_not_cross_pollute_unrelated_rel_table`: `InvalidArgument("relationship type '...' not found in catalog for labels (...)")`

All pass with the fix applied.

## Test plan
- [x] `regression_436_dangling_edges.rs` (5 new tests) — pass with fix, fail (3/5) against pre-fix commit with the exact reported symptoms
- [x] `delete_edge.rs`, `spa_216_delete_node.rs`, `spa_186_csr_nodeid.rs`, `spa_201_csr_backward.rs` — pass
- [x] `regression_486_edge_direction.rs`, `regression_487_488_edge_direction.rs` — pass
- [x] `spa_222_csr_lazy_load.rs` (slow suite) — pass
- [x] `cargo fmt`, `cargo clippy -p sparrowdb -p sparrowdb-execution --all-targets` — clean
- [x] `cargo check -p sparrowdb` — clean
- [ ] Full `cargo test -p sparrowdb` — running in background as extra diligence; will report if it surfaces anything

### DIFF
--- a/crates/sparrowdb/src/write_tx.rs
+++ b/crates/sparrowdb/src/write_tx.rs
@@ -456,45 +456,72 @@ impl WriteTx {
     ///
     /// [`commit`]: WriteTx::commit
     pub fn delete_node(&mut self, node_id: NodeId) -> crate::Result<()> {
-        // SPA-185: check ALL per-type delta logs for attached edges, not just
-        // the hardcoded RelTableId(0).  Always include table-0 so that any
-        // edges written before the catalog had entries are still detected.
-        let rel_entries = self.catalog.list_rel_table_ids();
-        let mut rel_ids_to_check: Vec<u32> =
-            rel_entries.iter().map(|(id, _, _, _)| *id as u32).collect();
-        // Always include the legacy table-0 slot.  If it is already in the
-        // catalog list this dedup prevents a double-read.
-        if !rel_ids_to_check.contains(&0u32) {
-            rel_ids_to_check.push(0u32);
-        }
-        for rel_id in rel_ids_to_check {
-            let store = EdgeStore::open(&self.inner.path, RelTableId(rel_id));
+        // SPA-185 / #436: check ALL per-type delta logs and CSR files for
+        // attached edges, not just the hardcoded RelTableId(0).
+        //
+        // CSR files are indexed by label-relative *slot*, not by the packed
+        // `NodeId` (which has the label folded into the upper 32 bits, see
+        // `NodeId` encoding). Indexing with the full `node_id.0` overflows
+        // `CsrForward::n_nodes`/`CsrBackward::n_nodes` for almost any label
+        // other than label 0, so `neighbors`/`predecessors` silently returned
+        // `&[]` and this check never fired for checkpointed edges — a node
+        // with a checkpointed edge could be deleted, leaving that edge
+        // dangling (#436). Strip the label bits before indexing.
+        //
+        // Each rel table's CSR is only meaningful for node_id when its label
+        // actually matches that table's declared src label (forward) or dst
+        // label (backward); a table for an unrelated label pair may have a
+        // numerically coincidental slot that is not actually node_id at all.
+        let node_slot = node_id.0 & 0xFFFF_FFFF;
+        let node_label_id = (node_id.0 >> 32) as u16;
 
-            // Check delta log (un-checkpointed edges).
-            if let Ok(ref s) = store {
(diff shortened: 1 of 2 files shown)


ryaker:
@coderabbitai review

Issue #459, opened 2026-08-02
CRITICAL: MATCH (n:L) RETURN hybrid_search(...) returns Null on a HEALTHY index — chunked pipeline never dispatches it


ryaker:
## The most natural way to write a hybrid search over matched nodes silently returns nothing

Measured against a **healthy** 5-vector index:

| query | rows | values |
|---|---|---|
| `RETURN hybrid_search(...)` | 1 | `["List"]` ✅ |
| `MATCH (d:Doc) RETURN hybrid_search(...)` | 5 | `["Null","Null","Null","Null","Null"]` ❌ |
| `MATCH (d:Doc) WHERE ... RETURN hybrid_search(...)` | 5 | all `Null` ❌ |
| `UNWIND [1..5] AS i RETURN hybrid_search(...)` | 5 | all `Null` ❌ |
| `MATCH (d:Doc) WITH d RETURN hybrid_search(...)` | 5 | `["List" × 5]` ✅ |

No error, no warning. Just `Null`.

## Cause

The chunked pipeline serves those scans and never routes `Expr::FnCall("hybrid_search")` through `eval_expr_graph`, so evaluation falls through to the generic evaluator, which has no implementation and yields `Null`.

**This is the same class as #415/#416** — the chunked pipeline bypassing a feature that works fine on the path it bypasses. Per [reference_chunked_pipeline_always_on], the chunked pipeline is live for every `db.execute()` because of `.with_chunked_pipeline()` at `db.rs:243`, despite the flag defaulting to false. So this is not an opt-in edge case; it is the default path.

The `WITH d` variant working is the tell: inserting a `WITH` forces a different execution path, and that path dispatches correctly.

## Why this is worse than it looks

`Null` is also the observable for a **damaged** index once #458 lands (deliberately — degrading to FTS-only was rejected because it collides with "index legitimately absent", the confusion #445 exists to kill). So from a consumer's seat, `hybrid_search returned null` will have two completely different causes:

- the index is damaged, and
- you wrote your query with `MATCH ... RETURN` instead of `MATCH ... WITH ... RETURN`, on a perfectly healthy store.

Fixing this is what keeps the #458 signal meaningful.

## How it was found — worth recording

[...]

Issue #369, opened 2026-03-29
bug: hop query RETURN with AS aliases always returns Null (projection uses alias not expression)


ryaker:
## Summary

`MATCH (n)-[r]->(m) RETURN n.name AS from, type(r) AS rel, m.name AS to` returns all `Null` values even though the edges and nodes exist.

## Root Cause

`project_hop_row` resolves values by pattern-matching the column name string:
- `"n.name"` → splits on `.` → looks up property ✅
- `"from"` (aliased from `n.name AS from`) → no dot, no match → `Value::Null` ❌

The function receives `column_names: &[String]` which contains aliases after `AS`, not the original expression strings. Any column using `AS alias` in a hop query returns `Null`.

## Reproduction

```cypher
-- All nulls (aliases)
MATCH (n)-[r]->(m) RETURN n.name AS from, type(r) AS rel, m.name AS to

-- Works (no aliases)  
MATCH (n)-[r]->(m) RETURN n.name, type(r), m.name
```

## Impact

Every demo-quality query uses AS aliases for readable column names. This makes hop queries effectively unusable for presentation.

## Fix

`project_hop_row` (and related projection functions) need to receive `&[ReturnItem]` (the original AST items) alongside column_names, and use `item.expr` for value resolution while using `column_names` for output headers. The scan path already does this correctly — it evaluates `item.expr` directly.

PR #514, opened 2026-08-15
fix(execution): projection resolves RETURN columns by AST expr, not display name (closes #444, #430)

### LINKED ISSUE #444: RETURN n.prop AS alias silently returns NULL unless alias == prop — silent wrong data on the most common query shape
## Severity

Aliased property projection — `RETURN n.name AS personName` — returns **NULL**. No error, no warning. It works only when the alias happens to equal the property name.

Aliasing is idiomatic Cypher and is what LLM-generated queries emit constantly, so this silently corrupts results on one of the most common query shapes in the language.

## Reproduction

Verified on `main` (`3205245`) with a standalone binary against a fresh DB:

```rust
db.execute("CREATE (n:Person {name: 'Alice', age: 30})")
```

| Query | `columns` | `rows` |
|---|---|---|
| `MATCH (n:Person) RETURN n.name` | `["n.name"]` | `[["Alice"]]` ✅ |
| `MATCH (n:Person) RETURN n.name AS name` | `["name"]` | `[["Alice"]]` ✅ |
| `MATCH (n:Person) RETURN n.name AS personName` | `["personName"]` | **`[[Null]]`** ❌ |
| `MATCH (n:Person) RETURN n.age AS years` | `["years"]` | **`[[Null]]`** ❌ |
| `MATCH (n:Person) RETURN n.name AS a, n.age AS b` | `["a","b"]` | **`[[Null, Null]]`** ❌ |

The column headers are correct. Only the values are wrong.

## Likely cause

The projection appears to resolve values by looking up the **output column name** rather than the source expression. When the alias matches the property name the lookup coincidentally succeeds; otherwise it misses and yields NULL. Property names are stored as `col_<fnv1a32(name)>` keys, so a lookup keyed on the alias string will not find `col_<hash("name")>` when the alias is `personName`.

Worth checking `extract_return_column_names` and the row-projection path in `crates/sparrowdb-execution/src/engine/` — the same place that builds `columns` from the alias.

## Why this matters beyond correctness

Found while building `langchain-sparrowdb` (WS3). `GraphCypherQAChain` has an LLM generate Cypher, and generated Cypher aliases almost everything — `RETURN p.name AS name, c.title AS company`. Every such query returns structurally valid rows full of NULLs, and the LLM then answers confidently from empty data. There is no failure mode the caller can detect.

This is the single highest-impact dialect gap for any generated-Cypher consumer.

## Related

Discovered alongside several other gaps documented empirically in `langchain-sparrowdb`'s `DIALECT.md`. Others worth separate triage:

- `MATCH (a)-[r]->(b) SET r.prop = ...` — relationship-property SET rejected outright
- `MERGE` with inline relationship properties **silently drops them** (`CREATE` does not)
- `keys(n)` always returns the literal string `"[]"`
- `COUNT(n)` is stale after delete while `MAX()` is not
- bare `RETURN n` exposes internal `col_<hash>` keys rather than property names
- `MATCH p = shortestPath(...)` (path-variable form) is a parse error

ryaker:
## Summary

Both #444 and #430 shared one root cause: the fast (non-eval) projection paths resolved a RETURN column's stored property by re-parsing the *output* column name — the alias, when one is present — instead of dispatching on the underlying `ReturnItem.expr`. Confirmed as one shared mechanism, not two independent bugs.

- **#444** — `collect_col_ids_from_columns` / `project_row` hashed the alias string as if it were the property name, so `RETURN n.name AS personName` hashed `col_id_of("personName")` — a column that never existed — instead of `name`'s real column, and silently projected `Null`. Replaced with `collect_col_ids_from_return_items` / a rewritten `project_row`, both dispatching on `item.expr` — the same pattern `project_hop_row` already established for the hop path when #369 was fixed. The identical anti-pattern also existed inline in `try_degree_sort_fastpath` (SPA-272 degree-cache fast path) and is fixed there too.
- **#430** — `execute_pipeline_match_hop` (the executor for a MATCH stage following `WITH`) only ever read `pat.rels[0]` / `pat.nodes[0..2]` and silently dropped every hop after the first. A 2+-hop pattern after `WITH` therefore bound only the first destination node — later RETURN variables projected `Null`, or worse, the query fabricated a row for a path that does not exist in the graph at all. Rewritten to walk every hop in `pat.rels` in turn, threading the accumulated binding map forward the same way `execute_n_hop` already does for the leading MATCH.

## Testing

New file: `crates/sparrowdb/tests/regression_444_430_null_projection.rs` — e2e against a real `GraphDb` on a `tempfile::TempDir`. Covers: alias == prop, alias != prop, multiple aliased columns in one RETURN, alias on an expression (`toUpper(...)`) rather than a bare property, the degree-fastpath aliasing gap, 1/2/3-hop after `WITH`, a post-WITH path that genuinely does not exist (must be `[]`, not a fabricated `[[Null]]` row), and a control where the property is genuinely absent (must stay `Null`, not become an error).

All expected values derived by hand from the fixtures. All 6 alias/hop-dependent tests confirmed to **fail against the pre-fix parent commit** with the predicted symptom before the source fix was applied — verified by stashing the source changes and re-running:

```
alias_different_from_property_name_resolves_correctly: left: [[Null]]  right: [[String("Alice")]]
multiple_aliased_columns_all_resolve:                  left: [[Null, Null]]  right: [[String("Alice"), Int64(30)]]
degree_fastpath_with_aliased_property_column:          left: [[Null]]  right: [[Int64(1)]]
three_hop_after_with_binds_fourth_node:                left: [[Null]]  right: [[String("D")]]
two_hop_after_with_binds_third_node:                   left: [[String("B"), Null]]  right: [[String("B"), String("C")]]
post_with_nonexistent_path_yields_empty_not_null_row:   left: 1 row [[Null]]  right: 0 rows  (the fabricated-row symptom)
```

`$CARGO fmt -- --check` and `$CARGO clippy -p sparrowdb -p sparrowdb-execution --all-targets --keep-going` both clean. `cargo check --workspace --exclude sparrowdb-ruby` clean.

Ran and passed (all green, no regressions): `regression_444_430_null_projection` (12), `regression_421_varlen_plus_hop` (15), `regression_369` (3), `regression_355` (3), `regression_467` (11), `regression_477_fts_where` (10), `spa_130_with_clause` (6), `spa_131_optional_match` (10), `spa_134_multi_clause` (13), `spa_138_case_when`, `spa_140_143_functions` (27), `spa_151_kms_query_validation`, `spa_155_unwind_param`, `spa_192_match_no_label` (5), `spa_194_count_node_var` (3), `spa_196_id_function` (4), `spa_197_count_label_fastpath`, `spa_207_labels_function` (4), `spa_213_return_node_var` (3), `spa_237_unwind_match`, `spa_242_count_rel_var`, `spa_251_text_search_index`, `spa_261_edge_props_perf`, `spa_263_two_hop_agg`, `spa_263_two_hop_null`, `spa_264_boolean_props`, `spa_265_backtick_escaping`, `spa_272_q7_count_fastpath`, `spa_272_q7_cypher_wiring` (5), `spa_289_multi_label` (10 + 2 pre-existing ignores), `spa_299_chunked_pipeline` (23), `spa_299_phase4_parity`, `spa_415_unwind_match_mutate`, `spa_98_wal_encryption`, `spa_aggregation`, `spa_collect_agg`, `spa_list_predicates`, `uc7_unwind`, `vector_index`, `acceptance`, `call_subquery`, `cypher_range_function_test`, `fts_index`, `gap_10_parameterized_queries`, `hybrid_search`, `mcp_cypher_templates`, `regression_363`, `regression_372`, `regression_406`, `regression_456_load_is_not_destructive`, `regression_462_fts_open_swallow`, `regression_475_473_null_coercion`, `regression_485_stale_count`.

Rebased onto current `origin/main` (`1a3b506`) before pushing — fast-forward only, no conflicts (the only new commit since my base was a CI-only change; sibling #459/#436 work had not landed on `origin/main` at push time).

## Could not verify

- **`regression_real_world.rs`: `knowledge_graph_facts_count` and `social_graph_company_count` hang** (60s+ timeouts hit, single-threaded, `--nocapture`). **Confirmed pre-existing and unrelated to this change** — reproduced identically by stashing this PR's source changes and re-running against unmodified `origin/main`. Not caused by this fix; left as-is and flagged for separate triage rather than fixed here (out of scope for #444/#430, and I didn't want to guess at a root cause for an unrelated hang under time pressure).
- Did not add coverage for `SET`/`MERGE`/mutation-path projection using the same `project_row`/`collect_col_ids_from_columns` machinery beyond what the existing suite already exercises — the fix is general (dispatches on `item.expr` for every caller), and the full targeted suite above passed, but I did not write new mutation-specific alias tests.
- Did not chase whether `execute_pipeline_match_hop`'s inline prop filters on hop 2+ can reference chain-local variables bound at hop 1 (e.g. `{x: b.y}` where `b` is itself mid-chain) — I deliberately kept `matches_prop_filter_with_binding` scoped to the upstream `binding` map only, matching pre-existing single-hop semantics exactly, rather than introducing new cross-hop filter semantics untested by any existing case.

## Test plan

- [x] `regression_444_430_null_projection` — 12/12, and 6/12 confirmed to fail pre-fix
- [x] Full targeted projection/pipeline/WITH/alias suite — all green
- [x] `cargo fmt --check`, `cargo clippy --keep-going` clean
- [x] `cargo check --workspace --exclude sparrowdb-ruby` clean
- [x] Rebased onto current `origin/main`, no conflicts

### DIFF
--- a/crates/sparrowdb-execution/src/engine/mod.rs
+++ b/crates/sparrowdb-execution/src/engine/mod.rs
@@ -2087,14 +2087,27 @@ fn prop_name_to_col_id(name: &str) -> u32 {
     col_id_of(name)
 }
 
-fn collect_col_ids_from_columns(column_names: &[String]) -> Vec<u32> {
+/// Collect column IDs referenced by property-access RETURN items.
+///
+/// Dispatches on the AST expression (`item.expr`), not on the (possibly
+/// aliased) output column name. `collect_col_ids_from_columns` — the function
+/// this replaces — parsed the *display* name (`RETURN n.name AS personName`
+/// produced `"personName"`) as if it encoded `var.prop`, so the property to
+/// fetch from storage was hashed from the alias instead of `name`. That
+/// silently omitted the real column from the read, which is the root cause of
+/// #444 (aliased projection returns Null) and part of #430's phantom rows.
+///
+/// See `project_row` for the corresponding projection-time fix, and
+/// `collect_col_ids_for_var_from_items` (#369) for the same item-based
+/// pattern already established on the hop/traversal paths.
+fn collect_col_ids_from_return_items(items: &[ReturnItem]) -> Vec<u32> {
     let mut ids = Vec::new();
-    for name in column_names {
-        // name could be "var.col_N" or "col_N"
-        let prop = name.split('.').next_back().unwrap_or(name.as_str());
-        let col_id = prop_name_to_col_id(prop);
-        if !ids.contains(&col_id) {
-            ids.push(col_id);
+    for item in items {
+        if let Expr::PropAccess { prop, .. } = &item.expr {
+            let col_id = prop_name_to_col_id(prop);
+            if !ids.contains(&col_id) {
+                ids.push(col_id);
+            }
         }
     }
     ids
@@ -2639,50 +2652,41 @@ fn eval_expr(expr: &Expr, vals: &HashMap<String, Value>) -> Value {
     }
 }
 
+/// Project a scanned node's properties into RETURN columns.
+///
+/// Dispatches on the AST expression (`items[i].expr`), not on the (possibly
+/// aliased) output column name — this fixes #444: `RETURN n.name AS
+/// personName` must resolve `n.name`, not `"personName"`. The old
+/// implementation split the *output* column string on `.` and hashed
+/// whatever followed, which only happened to work when the alias equalled
+/// the property name.
+///
+/// Every call site only reaches `project_row` once `id(var)`, `labels(var)`,
+/// bare `var`, and every other non-aggregate `FnCall` have already been
+/// routed to the eval path (`needs_node_ref_in_return` /
+/// `expr_needs_eval_path`, checked by every caller before choosing between
+/// this fast path and the eval path). So every item here is guaranteed to be
+/// a `PropAccess`; the `_ => Value::Null` arm is a defensive fallback, not a
+/// reachable case.
 fn project_row(
     props: &[(u32, u64)],
-    column_names: &[String],
-    _col_ids: &[u32],
-    // Variable name for the scanned node (e.g. "n"), used for labels(n) columns.
+    items: &[ReturnItem],
+    // Variable name for the scanned node, used to disambiguate PropAccess.
     var_name: &str,
-    // Primary label for the scanned node, used for labels(n) columns.
-    node_label: &str,
     store: &NodeStore,
-    // NodeId of the scanned node, used for id(var) columns.
-    node_id: Option<NodeId>,
 ) -> Vec<Value> {
-    column_names
+    items
         .iter()
-        .map(|col_name| {
-            // Handle id(var) column — returns the node's integer id.
-            if let Some(inner) = col_name
-                .strip_prefix("id(")
-                .and_then(|s| s.strip_suffix(')'))
-            {
-                if inner == var_name {
-                    if let Some(nid) = node_id {
-                        return Value::Int64(nid.0 as i64);
-                    }
-                }
-                return Value::Null;
+        .map(|item| match &item.expr {
+            Expr::PropAccess { var, prop } if var.as_str() == var_name => {
+                let col_id = prop_name_to_col_id(prop);
+                props
+                    .iter()
+                    .find(|(c, _)| *c == col_id)
+                    .map(|(_, v)| decode_raw_val(*v, store))
+                    .unwrap_or(Value::Null)
             }
-            // Handle labels(var) column.
-            if let Some(inner) = col_name
-                .strip_prefix("labels(")
-                .and_then(|s| s.strip_suffix(')'))
-            {
-                if inner == var_name && !node_label.is_empty() {
-                    return Value::List(vec![Value::String(node_label.to_string())]);
-                }
-                return Value::Null;
-            }
-            let prop = col_name.split('.').next_back().unwrap_or(col_name.as_str());
-            let col_id = prop_name_to_col_id(prop);
-            props
-                .iter()
-                .find(|(c, _)| *c == col_id)
-                .map(|(_, v)| decode_raw_val(*v, store))
-                .unwrap_or(Value::Null)
+            _ => Value::Null,
         })
         .collect()
 }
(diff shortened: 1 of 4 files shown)


ryaker:
@coderabbitai review

ryaker:
Independent verification (team-lead), run in a separate worktree with its own `CARGO_TARGET_DIR` so it shares nothing with the author's build.

**The risk I went after** was the one hedged in the PR description: `project_row` loses its string-based handling of `id(var)` and `labels(var)`, and its new fallback arm is `_ => Value::Null`. If anything reaches it that isn't a matching `PropAccess`, this fix silently reintroduces the exact bug class it exists to close. "Verified empirically via spa_196/207/213 passing" is not enough to retire that risk — passing tests show no *covered* case regressed, not that the arm is unreachable.

**It is actually safe by construction, which is stronger than the PR claims.** `needs_node_ref_in_return` routes to the eval path when `expr_needs_eval_path` is true, and that function returns `true` for *every* non-aggregate `FnCall` — with a comment already on `main` saying exactly why: "we include id/labels/type here even though they are special-cased in eval_expr, because the fast project_row path cannot handle them at all." So `id(n)`/`labels(n)` cannot reach `project_row`. The deleted code was already dead. Worth stating in those terms in the PR body, because the structural argument is what makes the deletion reviewable.

Confirmed by probe rather than by reading (6/6 pass on this branch): `id(n)`, `labels(n)`, `id(n) AS ident`, `labels(n) AS lbls`, and `n.name AS nm, labels(n) AS lbls` in one row.

**The fix reproduces and resolves #444 independently of the author's tests.** Same probe against `main` @ 7fbdc51 vs this branch:

```
main:   MATCH (n:Widget) RETURN n.name AS nm, 42  ->  [[Null, Null]]
branch: MATCH (n:Widget) RETURN n.name AS nm, 42  ->  [[String("w1"), Null]]
```

**Two notes for the record, neither blocking:**

1. The residual `Null` above is a *different, pre-existing* defect — every literal or arithmetic expression in a RETURN after a MATCH returns Null on `main`. Filed as #515 with a full table. Not caused by this PR, and I confirmed it on `main` before filing.

2. The reported "hang" in `regression_real_world` is not a hang. CI runs `social_graph_company_count` in ~196s and the suite in ~240s, green. It prints nothing while running and this box is loaded with concurrent agent builds. Nothing to triage.

CI green except benchmarks/claude-review still running. Holding merge for CodeRabbit, which is rate-limited repo-wide right now — this diff is the one worth spending that quota on.

Issue #502, opened 2026-08-15
PANIC: a bare CALL { <mutation> } reaches unreachable!() and crashes the process


ryaker:
A top-level `CALL { }` containing a mutating statement **panics**. For an embedded database that is the host process, not a query error.

## Reproduction on `main` (0.1.24)

```cypher
CALL { CREATE (:M1 {z:1}) }
```

```
thread 'call_shapes' panicked at crates/sparrowdb/src/db.rs:566:22
```

Only the **bare, top-level** form panics. Every pipeline form fails closed correctly:

| query | result |
|---|---|
| **`CALL { CREATE (:M1 {z:1}) }`** | ***PANIC*** |
| `MATCH (s:Seed) WITH s CALL { CREATE (:M2 {z:2}) } RETURN s.id` | `ERR CALL { }: unsupported or mutating statement` |
| `… CALL { CREATE (:M3 {z:3}) RETURN 1 AS x } …` | same clean error |
| `… CALL { MERGE (:M4 {z:4}) } …` | same clean error |
| `… CALL { MATCH (n:Seed) SET n.id = 99 } …` | same clean error |

## Cause: a classification/dispatch mismatch

`Engine::is_mutation` counts `Statement::CallSubquery` as a mutation (it recurses into the inner statement). So a `CALL { CREATE … }` takes the mutation branch at `db.rs:556`:

```rust
if Engine::is_mutation(&bound.inner) {
    match bound.inner {
        Statement::Merge(..)              => …,
        Statement::MatchMergeRel(..)      => …,
        Statement::MatchMutate(..)        => …,
        Statement::UnwindMatchMutate(..)  => …,
        Statement::MatchCreate(..)        => …,
        Statement::Create(..)             => …,
        _ => unreachable!(),          // <- CallSubquery lands here
    }
```

**`is_mutation` says yes; the dispatch has no arm for it.** The two lists have to agree and nothing enforces that they do — the same drift-prone shape that produced #478, where `is_mutation` was missing `Union` entirely.

## Why this matters more than the wrong answer it replaces

An `unreachable!()` reachable from user input is a crash. In an embedded engine the caller *is* the process — a malformed query from an agent or an end user takes down the host. It is also unrecoverable and untestable from the caller's side, unlike an `Err`.

## Suggested fix

Two parts:

[...]

PR #516, opened 2026-08-15
fix(execution): hybrid_search()/UNWIND RETURN no longer silently return Null (closes #459)

### LINKED ISSUE #459: CRITICAL: MATCH (n:L) RETURN hybrid_search(...) returns Null on a HEALTHY index — chunked pipeline never dispatches it

ryaker:
## Summary

`MATCH ... RETURN hybrid_search(...)` (and the `WHERE`/`UNWIND` variants) returned `Value::Null` against a genuinely **healthy** index — issue #459. **Not the chunked pipeline**, contrary to the issue title's working theory: `can_use_chunked_pipeline` already excludes any non-aggregate `FnCall` in RETURN via `expr_needs_eval_path`, so the chunked scan path is never reached for this query shape. Two independent routing bugs instead:

1. **`aggregate_rows_graph` (`engine/expr.rs`)** only recurses into the engine-aware `eval_expr_graph` when `expr_needs_graph()` (`engine/mod.rs`) returns `true` for a RETURN item. That function checked for `ShortestPath`/`ExistsSubquery`/`CaseWhen` but never `Expr::FnCall`, so a bare `hybrid_search(...)` fell through to the free `aggregate_rows()` helper's non-aggregate "plain projection" branch, which evaluates every item with the **non-engine-aware** `eval_expr()`. That function's `FnCall` arm dispatches to `functions::dispatch_function`, which has never heard of `hybrid_search`/`full_text_search`/`bm25_score` (those three are special-cased only inside `Engine::eval_expr_graph`) — so it silently returns `Null` for an unrelated reason: unknown function name. Explains the `MATCH ... RETURN` and `MATCH ... WHERE ... RETURN` rows.

2. **`Engine::execute_unwind` (`engine/mutation.rs`)** is a separate code path — bare `UNWIND ... RETURN` parses to `Statement::Unwind`, not `Statement::Pipeline`. It hand-rolled RETURN projection: a bare `Expr::Var(alias)` resolved to the unwound value, and *every other expression shape* — `hybrid_search(...)`, arithmetic, anything — was hardcoded to `Value::Null`. Independent bug, same symptom, explains the `UNWIND ... RETURN` row.

## Fix

- `expr_needs_graph()`: added an `Expr::FnCall` arm matching the same three names `eval_expr_graph` already special-cases. This single change fixes the RETURN projection across every scan variant that shares `aggregate_rows_graph` (single-label scan, multi-label scan, secondary-label scan, one-hop, ...).
- `execute_unwind()`: replaced the hand-rolled `Var`-only check with a proper row-value map (alias binding + `$params`) evaluated via `self.eval_expr_graph(...)` for every RETURN item. Preserves "unrelated variable → `Null`" (an absent map key still evaluates to `Null`) while making `hybrid_search` and any other expression resolve correctly.

## What did NOT change

`Engine::eval_hybrid_search` itself — the #456/#458/#465 absent-vs-damaged guard — is untouched. Verified with a corrupt-index e2e test through the exact `MATCH ... RETURN` shape this fix changes: still returns `Null`, not a fabricated result.

## Testing

New: `crates/sparrowdb/tests/regression_459_hybrid_search_return.rs` — 6 e2e tests against a real `GraphDb`/`tempfile::TempDir`, every expected value hand-derived from the fixture (never captured from a run):

- `match_return_hybrid_search_healthy_index_returns_hits`
- `match_where_return_hybrid_search_healthy_index_returns_hits`
- `unwind_return_hybrid_search_healthy_index_returns_hits`
- `unwind_return_bare_alias_still_works` (no-regression baseline — passed pre-fix too)
- `match_return_hybrid_search_no_match_returns_empty_list_not_null` (healthy-but-empty index → empty `List`, never `Null`)
- `match_return_hybrid_search_corrupt_index_still_returns_null` (guard preserved — reuses the `truncate_to_4_bytes` technique from `regression_456_load_is_not_destructive.rs`)

All 5 hit-producing tests confirmed to fail against the pre-fix commit with `expected List, got Null` (quoted in the commit message / session log).

Also re-ran clean: `hybrid_search.rs`, `fts_index.rs`, `regression_462_fts_open_swallow.rs`, `regression_456_load_is_not_destructive.rs`, `regression_477_fts_where.rs`, `vector_index.rs`, `vector_index_durability.rs`, `vector_index_load_errors.rs`, `uc7_unwind.rs`, `acceptance.rs`, `regression_475_473_null_coercion.rs`, `spa_134_multi_clause.rs`, plus the full `cargo test --workspace --exclude sparrowdb-ruby --no-fail-fast` (237 test binaries, 0 failures).

`cargo fmt --check` clean. `cargo clippy -p sparrowdb -p sparrowdb-execution --tests --keep-going -- -D warnings` clean.

Rebased onto current `origin/main` (through #512) before opening.

## Could not verify

- Whether some test file outside the ones I grepped for (`execute_unwind`/`"UNWIND ["`) exercises bare `UNWIND ... RETURN` with a non-alias `FnCall` in a way I didn't anticipate — the full workspace run (237 binaries, 0 failures) is the backstop here, but I did not enumerate every UNWIND-touching test by hand.
- Whether #514 (in flight per the team lead, touches `engine/mod.rs` in the same `expr_needs_graph`/`needs_node_ref_in_return` neighborhood) will conflict with this PR once merged — it had not landed on `origin/main` as of this rebase, so I could not test against it.

Closes #459

### DIFF
--- a/crates/sparrowdb-execution/src/engine/expr.rs
+++ b/crates/sparrowdb-execution/src/engine/expr.rs
@@ -1,6 +1,55 @@
 //! Auto-generated submodule — see engine/mod.rs for context.
 use super::*;
 
+/// Signature shared by every scalar function that can only be evaluated with
+/// engine (graph) context — `eval_full_text_search`, `eval_bm25_score`, and
+/// `eval_hybrid_search` all have this exact shape.
+type GraphOnlyFn = fn(&Engine, &[Expr], &HashMap<String, Value>) -> Value;
+
+/// The single source of truth for which scalar functions require
+/// `Engine::eval_expr_graph` rather than the free, non-engine `eval_expr`,
+/// and how to dispatch each one.
+///
+/// # #459: one table instead of two name lists
+///
+/// Before this table existed, `eval_expr_graph`'s `FnCall` arm and
+/// `expr_needs_graph` (`engine/mod.rs`) each kept their own copy of the same
+/// three names, with nothing keeping them in sync. When they agreed the
+/// system worked; the day someone added a fourth graph-only function to one
+/// list and not the other, `expr_needs_graph` would keep saying "no" for it,
+/// `aggregate_rows_graph` would keep delegating to the non-engine
+/// `aggregate_rows` → `eval_expr`, and the new function would silently
+/// return `Value::Null` against a perfectly healthy index — this issue,
+/// verbatim, for the next function. This project has hit the same shape
+/// twice already for a match/dispatch pair that can drift (`is_mutation` vs
+/// its dispatch match: #478 missing-from-classifier, #502
+/// missing-from-dispatch).
+///
+/// A table is used instead of a plain name list so the dispatch itself reads
+/// from the same data `is_graph_only_fn` checks against — adding a function
+/// here is the only step; both the "does this need graph context" check and
+/// the actual call site update together. See `eval_expr_graph`'s `FnCall`
+/// arm below and `regression_459_hybrid_search_return.rs` /
+/// `graph_only_fn_table_covers_every_dispatched_name` (this file's tests).
+///
+/// To add a new graph-only function: add its `eval_*` method to this `impl
+/// Engine` block, then add one row here. Nothing else needs to change.
+const GRAPH_ONLY_FNS: &[(&str, GraphOnlyFn)] = &[
+    ("full_text_search", Engine::eval_full_text_search),
+    ("bm25_score", Engine::eval_bm25_score),
+    ("hybrid_search", Engine::eval_hybrid_search),
+];
+
+/// Returns `true` when `name` (case-insensitive) names one of
+/// [`GRAPH_ONLY_FNS`] — a scalar function that can only be evaluated via
+/// `Engine::eval_expr_graph`. Used by `expr_needs_graph` (`engine/mod.rs`)
+/// to route `aggregate_rows_graph` correctly; see the module doc on
+/// `GRAPH_ONLY_FNS` for why this must not be a second, independent list.
+pub(crate) fn is_graph_only_fn(name: &str) -> bool {
+    let name_lc = name.to_ascii_lowercase();
+    GRAPH_ONLY_FNS.iter().any(|(n, _)| *n == name_lc)
+}
+
 /// Key of the process-global `hybrid_search` index cache: which database
 /// directory, which `(label, prop)` pair.
 type HybridCacheKey = (std::path::PathBuf, String, String);
@@ -665,12 +714,17 @@ impl Engine {
                 }
                 Value::Null
             }
-            Expr::FnCall { name, args } => match name.to_ascii_lowercase().as_str() {
-                "full_text_search" => self.eval_full_text_search(args, vals),
-                "bm25_score" => self.eval_bm25_score(args, vals),
-                "hybrid_search" => self.eval_hybrid_search(args, vals),
-                _ => eval_expr(expr, vals),
-            },
+            // Dispatch through the shared GRAPH_ONLY_FNS table (#459) rather
+            // than a hand-written match, so this call site and
+            // `is_graph_only_fn` can never disagree about which names route
+            // here.
+            Expr::FnCall { name, args } => {
+                let name_lc = name.to_ascii_lowercase();
+                match GRAPH_ONLY_FNS.iter().find(|(n, _)| *n == name_lc) {
+                    Some((_, f)) => f(self, args, vals),
+                    None => eval_expr(expr, vals),
+                }
+            }
             // #477: a bare `full_text_search(...)`/`bm25_score(...)` call is
             // routed above, but a threshold comparison like
             // `bm25_score(n.text, 'q') > 0.5` wraps the call in a `BinOp` —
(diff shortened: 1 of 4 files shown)


ryaker:
@coderabbitai review

ryaker:
Merging. Recording the review state plainly so the green checks aren't read as more than they are.

**No external reviewer saw this PR.** `claude-review` no-opped twice — 4 turns/$0.155 on the first run, 5 turns/$0.176 after the re-request — against 12 turns/$0.81 on #517 and 10 turns/$5.90 on #514 the same evening. CodeRabbit returned `Action not completed — review rate limited` to both attempts; its repo-wide quota has been exhausted all night. So the checkmarks here represent CI only, and my own review is the only one this change received.

What I actually verified, so that's on the record too:

- **The `GRAPH_ONLY_FNS` design**, which is the point of the change: the dispatcher now *reads the table* rather than keeping a parallel hand-written match, so adding a function updates routing and dispatch together, and `GraphOnlyFn` makes a wrong fn pointer a compile error. A shared name list would not have been enough — it leaves the dispatch as a separate match that can omit an entry.
- **The failure mode I expected the tests to miss**: a table entry written with any uppercase would match nothing, never dispatch, and return Null on a healthy index — #459 verbatim. The pin test catches it, because it builds its `Expr::FnCall` from the table's own names.
- **Composition with #514.** `git merge-base --is-ancestor 606d9a2 <head>` confirms this branch sits on top of #514, so this CI run *is* the composition test — and a better one than the throwaway worktree I used earlier, since it includes the Ruby job that cannot compile locally. Benchmarks green too, which was worth waiting for: the FnCall arm went from a compiler-optimised `match` to a linear table scan in an expression-evaluation hot path.

This retires a bug class this repo has hit three times — #478 (missing from classifier → silent no-write), #502 (missing from dispatch → panic), #459 (missing from routing check → Null on a healthy index). That is worth more than the instance fix.

PR #517, opened 2026-08-15
fix(npm): drop dishonest optionalDependencies, fix darwin-arm64 load bug

### LINKED ISSUE #481: npm packaging: @sparrowdb/* platform packages are NOT published — only linux-x64 and darwin-arm64 actually work
`sparrowdb@0.1.21` advertises five platforms and ships binaries for **two**. On the other three it fails at import with a module-resolution error.

## Evidence

`package.json` declares five `optionalDependencies`:

```
@sparrowdb/linux-x64-gnu, @sparrowdb/linux-arm64-gnu,
@sparrowdb/darwin-x64, @sparrowdb/darwin-arm64, @sparrowdb/win32-x64-msvc
```

**None of them exist on the registry.** Direct query:

```
https://registry.npmjs.org/@sparrowdb%2Flinux-x64-gnu   -> {"error":"Not found"}
https://registry.npmjs.org/@sparrowdb%2Fdarwin-arm64    -> {"error":"Not found"}
https://registry.npmjs.org/@sparrowdb%2Fwin32-x64-msvc  -> {"error":"Not found"}
```

They are `optionalDependencies`, so npm skips them **silently** — no warning, no install failure.

The package works anyway because the tarball **bundles binaries directly** (3.0 MB):

```
package/sparrowdb.darwin-arm64.node    2.0M
package/sparrowdb.linux-x64-gnu.node   2.3M
package/sparrowdb.node                 2.3M
```

`index.js` tries the optional dependency first, then falls back to a local `sparrowdb.node`. So installs succeed on **linux-x64** and **darwin-arm64** only. `linux-arm64`, `darwin-x64`, and `win32-x64` resolve nothing and throw.

## Why this needs correcting explicitly

The Zora integration report (#480) reads the `optionalDependencies` list and concludes *"Platform coverage — better than it looks... Good coverage."* That is the reasonable inference from the manifest, and it is wrong. A consumer planning a deployment on ARM Linux or Windows would find out at runtime.

This matters more than a normal packaging bug because **the manifest is the only thing anyone can inspect before installing**, and it currently overstates support 5-to-2.

## What to fix

1. **Decide the strategy and make the manifest match it.** Either publish the five `@sparrowdb/*` packages from CI (the napi-rs convention, keeps the main tarball small), or drop the `optionalDependencies` entirely and be honest that binaries are bundled. The present state is the worst of both: it claims the first while doing the second.
2. **If keeping the bundled approach**, add the three missing binaries or trim the advertised list. A 3 MB tarball for two platforms becomes ~7 MB for five, which is the real cost of the simpler approach.
3. **Make the unsupported-platform failure catchable and specific.** `index.js` already builds a decent message naming the platform and the supported set — worth ensuring that path is what actually fires rather than a bare `MODULE_NOT_FOUND` from the optional-dependency `require`.
4. **No musl build** (already noted in #480) — Alpine is unsupported on every architecture, which is worth stating in the README rather than discovered in a Dockerfile.

Found while investigating what needs publishing to get 0.1.22's `executeWithParams` to an external consumer.

ryaker:
## Summary

Closes #481.

`sparrowdb@0.1.25` still declares five `@sparrowdb/<platform>` `optionalDependencies`; none of the five packages exist on npm (404 on all). Because they're optional, npm skips them silently — no install warning — so the manifest overstates platform support.

**The true severity is worse than #481 states, and worse than my first pass here framed it.** Verified by installing, not by reading the manifest — darwin-arm64 is broken on **every published version**, not "2 of 5 working, 3 broken":

```
0.1.20  BROKEN (ERR_DLOPEN_FAILED)
0.1.23  BROKEN
0.1.24  BROKEN
0.1.25  BROKEN
```

`npm install sparrowdb` has **never** worked on Apple Silicon. Actual state is **1 of 5 advertised platforms**, not 2 of 5. Root cause: CI's "generic fallback" step unconditionally copied the Linux binary to a platform-agnostic `sparrowdb.node`, and `index.js`'s dev-fallback step loaded that file by name with no platform check — so a fresh `npm install sparrowdb` on macOS ARM threw `ERR_DLOPEN_FAILED: slice is not valid mach-o file` trying to dlopen the bundled Linux `.so`.

## Decision

Per direction: drop `optionalDependencies` and be honest that binaries are bundled directly in the tarball, rather than build out the five-package napi-rs split-publish pipeline. The bundled approach is one moving part; splitting into five packages adds five more atomic-publish steps to a release pipeline that only became dual-registry (npm + crates.io) one day before this fix.

## Changes

- `npm/sparrowdb/package.json` / `package-lock.json` — remove `optionalDependencies`.
- `npm/sparrowdb/index.js` — replace the optional-dependency resolution step with a direct lookup of the platform-specific bundled filename (`sparrowdb.<platform>-<arch>.node`), checked *before* the generic dev-only fallback so it can never be shadowed by the wrong platform's binary. **Every** load attempt (bundled binary, dev-local `sparrowdb.node`, workspace target dir) is now wrapped in a `tryLoad()` helper — a file existing under the expected name is not proof it's valid; a failed `require()` falls through to the next candidate and, ultimately, to a single catchable `Error` listing what was tried and why, instead of letting a raw dlopen/ABI error escape uncaught. This is the actual mechanism that broke darwin-arm64: the bad binary loaded from a name that was `require`d unconditionally with no try/catch.
- `.github/workflows/release.yml` — stop publishing the generic `sparrowdb.node` fallback name (that's what caused the live darwin-arm64 break); drop the now-dead `optionalDependencies` version-sync step.
- `.github/workflows/ci.yml` / `npm/sparrowdb/package.json` — `npm test` now runs `test/*.test.js` (was hardcoded to a single file) so the new suite below actually executes in CI.
- `npm/sparrowdb/test/platform-resolution.test.js` (new) — regression coverage for the class of bug that shipped. `integration.test.js` only ever loads a dev-built `sparrowdb.node` copied onto the exact machine that built it, which is trivially always correct for that machine — it can't see a mismatch between a platform-specific bundled filename and its contents. That's why CI (`ubuntu-latest` only, testing only the generic dev-fallback name) never caught the darwin-arm64 break across five releases. The new suite builds an isolated copy of the package dir the way `npm publish` does — nothing but a binary under `sparrowdb.<platform>-<arch>.node` — and requires it in a fresh child process, asserting: (a) a correctly-bundled binary for the current platform loads and exposes the API, (b) a bad/wrong-platform binary under the expected name falls through to the catchable error rather than crashing uncaught, (c) an explicitly unsupported platform+arch throws the same catchable, specific error.
- `README.md` — document the two actually-bundled platforms, that musl/Alpine is unsupported on every architecture, and how to build from source elsewhere.

## Test plan

- [x] `npm pack --dry-run` against a locally-built darwin-arm64 binary — tarball lists exactly `sparrowdb.darwin-arm64.node`, matching the two-platform claim.
- [x] Forced the unsupported-platform path (`win32-x64`, `linux-arm64`, `darwin-x64`) — each throws the catchable `Error: sparrowdb: no prebuilt native module for <platform>...`, not a bare `MODULE_NOT_FOUND`/dlopen crash.
- [x] **Confirmed the new regression test fails against the pre-try/catch code** — checked out the prior commit's `index.js` (platform-specific lookup present, but no try/catch), ran `platform-resolution.test.js` against it, watched it fail with the exact reported symptom (`dlopen(...): slice is not valid mach-o file`), then restored the fix and watched it pass. Guards what it claims to guard.
- [x] `npm test` — 50/50 pass (was 47) on darwin-arm64 against a real compiled binary.
- [x] `npm run typecheck` passes.
- [x] Rebased onto current `origin/main` (#514 landed since the branch was cut) — clean rebase, retested green after.
- [x] `npx prettier --write` run on changed JS/JSON, then reverted: this repo has no `.prettierrc`, so Prettier defaults to double-quotes/semicolons against an established single-quote/no-semi codebase and wanted to reformat hundreds of unrelated lines in `README.md` and `integration.test.js`. Hand-applied only the functional diff in the existing style instead, to keep the PR reviewable. New files (`platform-resolution.test.js`) were hand-written in the same convention.

Not merging — opening/updating only, per instructions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -232,7 +232,7 @@ jobs:
           echo "Copied $ARTIFACT → npm/sparrowdb/sparrowdb.node"
 
       - name: Run Node.js integration tests
-        run: node --test npm/sparrowdb/test/integration.test.js
+        run: node --test npm/sparrowdb/test/*.test.js
 
   # ── Ruby bindings (compile check) ─────────────────────────────────────────
   ruby-bindings:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,16 +93,17 @@ jobs:
 
       - name: Assemble npm package
         run: |
-          # Copy native binaries into npm package directory
+          # Copy native binaries into npm package directory. index.js maps
+          # process.platform+arch to exactly these filenames — do NOT also
+          # write a platform-agnostic `sparrowdb.node` here. That name is
+          # reserved for local dev builds (see index.js); publishing one
+          # from CI made every install silently load whichever platform's
+          # binary happened to be copied last, regardless of the actual
+          # runtime platform (issue #481).
           cp artifacts/sparrowdb-node-linux-x64/libsparrowdb_node.so \
              npm/sparrowdb/sparrowdb.linux-x64-gnu.node 2>/dev/null || true
           cp artifacts/sparrowdb-node-darwin-arm64/libsparrowdb_node.dylib \
              npm/sparrowdb/sparrowdb.darwin-arm64.node 2>/dev/null || true
-
-          # Also create a generic fallback name
-          if [ -f npm/sparrowdb/sparrowdb.linux-x64-gnu.node ]; then
-            cp npm/sparrowdb/sparrowdb.linux-x64-gnu.node npm/sparrowdb/sparrowdb.node
-          fi
 
       - name: Sync version to package.json
         run: |
@@ -111,12 +112,6 @@ jobs:
           node -e "
             const pkg = require('./package.json');
             pkg.version = '$VERSION';
-            // Update optional dependency versions too
-            if (pkg.optionalDependencies) {
-              for (const dep of Object.keys(pkg.optionalDependencies)) {
-                pkg.optionalDependencies[dep] = '$VERSION';
-              }
-            }
             require('fs').writeFileSync('package.json', JSON.stringify(pkg, null, 2) + '\n');
           "
           echo "package.json version set to $VERSION"
--- a/README.md
+++ b/README.md
@@ -147,6 +147,24 @@ SparrowDB is *not* the right choice when:
 npm install sparrowdb
 ```
 
+The `sparrowdb` npm package bundles prebuilt native binaries directly in the tarball — there is no separate `@sparrowdb/<platform>` package to resolve.
+
+| Platform | Supported |
+|---|---|
+| macOS arm64 (Apple Silicon) | Yes |
+| Linux x64 (glibc) | Yes |
+| macOS x64 (Intel) | No — build from source |
+| Linux arm64 | No — build from source |
+| Windows x64 | No — build from source |
+| Linux x64/arm64 (musl / Alpine) | No — unsupported on every architecture |
+
+On an unsupported platform, `require('sparrowdb')` throws a clear error naming the platform and the supported list, rather than a raw module-resolution failure. To build from source on those platforms:
+
+```bash
+cargo build --release -p sparrowdb-node
+cp target/release/libsparrowdb_node.<so|dylib|dll> npm/sparrowdb/sparrowdb.node
+```
+
 ### Rust
 
 ```toml
--- a/npm/sparrowdb/index.js
+++ b/npm/sparrowdb/index.js
@@ -14,28 +14,48 @@ const PLATFORM_BINARIES = {
   'darwin-arm64': 'sparrowdb.darwin-arm64.node',
 }
 
+// A file existing under an expected name is not proof it's a valid,
+// loadable binary for this platform — issue #481 shipped for a year because
+// a mis-copied binary sat at a name that was `require`d unconditionally, and
+// the resulting dlopen/ABI error propagated straight out of loadNative()
+// uncaught instead of falling through to the clear message below. Every
+// candidate here is therefore tried, not just checked for existence.
+function tryLoad(path) {
+  try {
+    return { ok: true, value: require(path) }
+  } catch (err) {
+    return { ok: false, error: err }
+  }
+}
+
 function loadNative() {
-  // 1. Try the binary bundled for this exact platform+arch (production / npm install).
+  const attempts = []
+
+  // 1. The binary bundled for this exact platform+arch (production / npm install).
   const key = `${process.platform}-${process.arch}`
   const bundled = PLATFORM_BINARIES[key]
   if (bundled) {
     const bundledPath = join(__dirname, bundled)
     if (existsSync(bundledPath)) {
-      return require(bundledPath)
+      const result = tryLoad(bundledPath)
+      if (result.ok) return result.value
+      attempts.push(`${bundledPath}: ${result.error.message}`)
     }
   }
 
-  // 2. Try a locally compiled sparrowdb.node in the same directory
+  // 2. A locally compiled sparrowdb.node in the same directory
   //    (development: `cargo build --release && cp target/release/libsparrowdb_node.so npm/sparrowdb/sparrowdb.node`).
   //    This is a dev-only convenience — CI never publishes a file by this
   //    name, so it can't shadow the platform-specific binaries above with
   //    the wrong platform's build (see issue #481).
   const local = join(__dirname, 'sparrowdb.node')
   if (existsSync(local)) {
-    return require(local)
+    const result = tryLoad(local)
+    if (result.ok) return result.value
+    attempts.push(`${local}: ${result.error.message}`)
   }
 
-  // 3. Try the workspace target directory (useful during development without copying).
+  // 3. The workspace target directory (useful during development without copying).
   //    Node native addons must be loaded as .node files regardless of platform.
   //    Use `napi-cli` or manually rename the compiled artifact:
   //      Linux:   target/release/libsparrowdb_node.so  → sparrowdb.node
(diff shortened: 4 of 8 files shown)


ryaker:
**Verified independently against the live registry, and it is worse than this PR states — in a way that raises the priority.**

`npm install sparrowdb` on darwin-arm64 is broken on **every published version**, not just current. Tested each on this Mac mini (darwin/arm64, Node 22.22.0):

| version | result |
|---|---|
| 0.1.20 | BROKEN — `ERR_DLOPEN_FAILED` |
| 0.1.23 | BROKEN |
| 0.1.24 | BROKEN |
| 0.1.25 | BROKEN |

So this was never a regression — the npm package has **never** worked on Apple Silicon via a plain install. The real state is **1 of 5 advertised platforms works** (linux-x64), not 2 of 5 as #481 and I both had it.

**The full mechanism, confirmed by inspecting the installed tree:**

```
sparrowdb.darwin-arm64.node   Mach-O 64-bit arm64          <- correct, present, unused
sparrowdb.linux-x64-gnu.node  ELF 64-bit x86-64
sparrowdb.node                ELF 64-bit x86-64            <- byte-identical to linux-x64
```

Published `index.js` resolves in this order:
1. `require('@sparrowdb/darwin-arm64')` — never published, 404, throws, caught.
2. `existsSync(join(__dirname,'sparrowdb.node'))` → true → `require(local)` — **loads a Linux ELF on macOS.**

Step 2 has no try/catch around the `require`, so the dlopen error propagates uncaught and the carefully-written "no prebuilt native module for <platform>" message at the bottom of the file **never fires**. That is why users see a raw `slice is not valid mach-o file` rather than the actionable error — worth calling out, since fixing resolution order alone would leave that latent for the next platform.

The correct binary was in the tarball the whole time and loads perfectly:

```
$ node -e "require('./node_modules/sparrowdb/sparrowdb.darwin-arm64.node')"
DIRECT LOAD OK — exports: ReadTx, SparrowDB, WriteTx
```

**So the package is one line of resolution order away from working on macOS**, which is exactly what this PR does. Good catch finding it by actually installing rather than reading the manifest — that is the same mistake #481 documents the Zora report making, and it would have been easy to repeat.

Two requests before merge:

1. **Confirm the uncaught-`require` path is fixed, not just reordered.** Wrap the bundled-binary load so a corrupt or mismatched `.node` surfaces the catchable message instead of a raw dlopen crash. You verified the *unsupported*-platform path throws cleanly; this is the *supported-but-broken-binary* path, which is what actually bit us here.
2. **Note in the PR body that this fixes a live break for all macOS users**, not only a manifest-honesty issue. It changes the release urgency, and whoever reads the changelog needs that.

Holding merge for CI. Your prettier judgement call was right — hand-applying the functional diff instead of letting it reformat the file against a repo with no `.prettierrc` keeps this reviewable.

ryaker:
Both changes made — pushed as 0bd0bca (rebased onto current main, #514).

**1. Uncaught require, fixed.** `loadNative()` now wraps every load attempt (bundled binary, dev-local fallback, workspace target dir) in a `tryLoad()` helper. A failed `require()` — for any reason, not just a missing file — falls through to the next candidate and ultimately to a single catchable `Error` that lists what was tried and why each attempt failed, instead of letting a raw dlopen/ABI error propagate as the uncaught exception. This is the actual mechanism that broke darwin-arm64 across every release: the bad binary sat at a name that was `require`d unconditionally with no try/catch.

**2. Regression test added, and confirmed it would have caught this.** `test/platform-resolution.test.js` builds an isolated copy of the package dir the way `npm publish` does — nothing but a binary under `sparrowdb.<platform>-<arch>.node` — and requires it in a fresh child process. Three assertions: correctly-bundled binary for the current platform loads and exposes the API; a bad/wrong-platform binary under the expected name falls through to the catchable error instead of crashing uncaught; an explicitly unsupported platform+arch throws the same catchable error.

Per this repo's own regression-test rule, I checked it actually guards something: checked out the pre-fix `index.js` (platform-specific lookup present, but no try/catch — i.e. my first commit here), ran the new test against it, and it failed with the exact live symptom: `dlopen(...): slice is not valid mach-o file`. Restored the fix, reran, green. `npm test` now runs `test/*.test.js` (was hardcoded to one file) so this suite actually executes — in both `package.json` and `ci.yml`. 50/50 tests pass.

Confirmed your 1-of-5 finding matches what I'm seeing and updated the PR body with your version table plus the corrected framing (this is a live break on every published version, not a manifest-tidiness issue).

ryaker:
Both requested changes verified, end-to-end, on real Apple Silicon hardware — not by reading the diff.

I took the `index.js` from `0bd0bca`, dropped it in a clean directory alongside the genuine `sparrowdb.darwin-arm64.node` extracted from the **published 0.1.25 tarball**, and ran it on this Mac mini (darwin/arm64, Node 22.22.0):

**Case 1 — the real-world fix.** Correct binary present:
```
LOADED OK: SparrowDB, WriteTx, ReadTx
```
That is the exact scenario that fails with `ERR_DLOPEN_FAILED` on every published version.

**Case 2 — supported platform, wrong/corrupt binary** (Linux ELF placed at the darwin-arm64 filename, reproducing 0.1.25's actual on-disk state):
```
Error: sparrowdb: no prebuilt native module for darwin-arm64.
Supported platforms: linux-x64, darwin-arm64
Tried and failed to load:
  - .../sparrowdb.darwin-arm64.node: dlopen(...) (slice is not valid mach-o file)
```
Catchable `Error`, with the underlying dlopen message preserved in the detail block. This is the part I actually cared about: making failures catchable must not make a broken binary *silently* fall through to a misleading "no prebuilt module" — and it doesn't, because `attempts` carries the real cause into the final message.

One cosmetic note, non-blocking: the headline still reads "no prebuilt native module" in Case 2, when in fact one exists and is broken. The detail block immediately disambiguates, so I'd leave it rather than churn the string.

**The new test is at the right layer**, which was the whole point. It builds a temp directory mimicking the *installed* layout rather than exercising the dev tree, and runs each load in a child process so a dlopen crash yields a readable assertion instead of killing the runner. Nice catch on the `ci.yml` change from a hardcoded `integration.test.js` to a glob — without it the new file would have sat there never running, which is a small instance of the same class of bug this PR fixes.

Review state for the record: `claude-review` was substantive here (118 turns, $4.47) and found nothing. CodeRabbit is rate-limited repo-wide and did not review. Merging on CI + that review + the hands-on verification above.

Holding only for the checks on `0bd0bca`.

Issue #513, opened 2026-08-15
index.d.ts has drifted from its NAPI generator — a regen reintroduces already-fixed bugs


ryaker:
`npm/sparrowdb/index.d.ts` is NAPI-generated from `crates/sparrowdb-node/src/lib.rs`, but the checked-in copy has been **hand-edited repeatedly** and the two have drifted. A straight regen today reintroduces already-fixed bugs.

Surfaced while fixing #449, by running the actual regen procedure from `fd64625` and diffing against the checked-in file — not by reading.

## What a regen from current `main` does today

1. **Reintroduces the `SparrowDb` vs `SparrowDB` factory-casing bug.** That was hand-corrected in #405; the correction lives only in the checked-in artifact, not in the Rust source. A regen undoes it.
2. **Drops doc-comment prose** that landed in `lib.rs` for #441 and #462 and was never synced into the checked-in `.d.ts` — so the artifact is *also* behind its source in places.

So the two files disagree in **both directions**: some fixes exist only in the artifact, some documentation only in the source.

## Why this bit us twice in one day

#449 was a nested doc comment in `lib.rs` producing an unparseable `.d.ts`. Fixing only the artifact would have been undone by the next `napi build`. And when the Rust prose was then reworded, the checked-in `.d.ts` immediately stopped matching what the generator produces — **the artifact and its generator desynced again within the same session**, which is how this class of drift accumulates unnoticed.

Every hand-edit to a generated file is a fix with a countdown on it.

## What would actually fix it

A dedicated regen-and-reconcile pass:

1. Run the regen, diff against the checked-in file, and for each difference decide which side is right.
2. Move every hand-fix that should persist into the Rust source — the `SparrowDB` casing needs a napi annotation, not an edit to the output.
3. Add a CI check that **regenerates and asserts no diff**, so the artifact cannot drift from its source again.

That last step is the real deliverable. #449 added `npm run typecheck`, which catches a `.d.ts` that does not *parse* — it cannot catch one that parses but no longer matches its generator.

## Not urgent, but it will recur

Nothing is broken right now: after #449 the checked-in file parses, is correct, and matches the current `lib.rs` for the blocks that were touched. The risk is entirely in the next person running a regen and silently reverting #405, or hand-editing again and widening the gap.

Issue #475, opened 2026-08-09
SET n.prop = <null> silently writes Int64(0), destroying the previous value


ryaker:
Reachable on `main` today, in shipped 0.1.22:

```rust
db.execute("CREATE (:Item {id: 'a', v: 7})")?;
let mut p = HashMap::new();
p.insert("nv".to_string(), Value::Null);
db.execute_with_params("MATCH (n:Item {id: 'a'}) SET n.v = $nv", p)?;
// MATCH (n:Item) RETURN n.id, n.v  ->  [["a", Int64(0)]]
```

`v` was `7`. After setting it to null it is **`0`** — a value the caller never supplied, indistinguishable from a genuine zero on read, and the original value is gone.

## Cause

`exec_value_to_storage` (`crates/sparrowdb/src/helpers.rs:110-119`):

```rust
match v {
    EV::Int64(n)   => Value::Int64(*n),
    EV::Float64(f) => Value::Float(*f),
    EV::Bool(b)    => Value::Int64(if *b { 1 } else { 0 }),
    EV::String(s)  => Value::Bytes(s.as_bytes().to_vec()),
    _              => Value::Int64(0),   // Null, List, Map, Vector, NodeRef, EdgeRef
}
```

The catch-all swallows six variants. `Null`, `List`, `Map`, and `Vector` are all values a caller can realistically hold — a JSON-sourced parameter with a null field is the ordinary case, not an exotic one.

## Severity

This is unrecoverable after the fact. `0` is a legal value, so nothing downstream can tell a coerced null from a real zero — no error, no warning, no way to audit which rows were affected. The write reports success.

## Relationship to other issues

- Sibling of **#473**, which is the same coercion on the **CREATE** path via `value_to_store_value` (`engine/mod.rs:1536`). Same bug, different function; both should be fixed together and probably share a helper.
- Distinct from **#467** (fail-open filters), which is about the wrong *rows* being matched. This is about the wrong *value* being written to correctly-matched rows.

## Fix direction

Reject rather than coerce. `SET n.prop = null` should either remove the property (matching how an absent property reads back) or return an error; a `List`/`Map`/`Vector` where a scalar is required should error rather than silently becoming `0`. Whatever is chosen, the catch-all arm must go — an unhandled variant should be a compile error via an exhaustive match, not a silent zero.

Found while adversarially reviewing #470. Confirmed present on `main` independently of that PR, and still present with #471 applied.

PR #520, opened 2026-08-15
fix(db): refuse vector writes to a quarantined-no-live-index pair (#451)

### LINKED ISSUE #451: Fail-closed open() is one-shot: after quarantine, the store reopens 'healthy' and silently drops vector writes again

ryaker:
## Summary

Closes #451.

**Lead finding: a real, previously-untested silent data-loss bug, distinct from what #451 describes.** Vector-write routing was keyed on property name alone (`vector_indexes.keys().any(|(_, prop_name)| prop_name == prop)`), ignoring which *label* the index actually belonged to. With a live index on `(Doc, embedding)` and no index at all on `(Note, embedding)`, writing a vector to `Note.embedding` was routed as "an index will handle this" — skipped the property write — and then the HNSW block looked up `(Note, embedding)`, found nothing, and did nothing. **No error, no property write, no index entry.** Confirmed pre-fix (see Testing): the call returns `Ok(QueryResult { columns: [], rows: [] })`. Fixed here as a consequence of threading `label` through the routing decision, which the #451 guard needed anyway.

**#451 itself, and a correction to its own framing:** after #442/#456's fail-closed `open()` quarantines a damaged HNSW index, the second `open()` finds no live `.bin` for that `(label, prop)` — indistinguishable from "never indexed." I verified the actual pre-fix behavior rather than trusting the issue text: **#475 (already merged) had already made every non-scalar property write loud**, so a vector write against the quarantined pair did *not* silently no-op — it already errored with `"property value is a list, which has no scalar storage representation..."`. The bug is that this message names neither the pair nor the quarantine, giving an operator (or KMSmcp) no way to tell "you configured this wrong" from "this used to work and your data is stuck in a quarantine file." **The fix's value is actionable vs. generic error, not silence vs. noise** — the issue's framing overstates what was silent; correcting that here so it doesn't get inherited as fact.

## What changed

**`crates/sparrowdb/src/{helpers,types,db}.rs`**
- `VectorIndexFailureKind` (`LiveUnserviceable` / `QuarantinedNoLiveIndex`) added as a field on `VectorIndexFailure`, so `vectorIndexHealth()`'s `active` list is machine-distinguishable, not just prose. Verified it clears: after `create_vector_index`, the artifact moves from `active` to `historical`.
- `DbInner.quarantined_vector_writes`: `(label, prop) → quarantine file paths`, seeded once at `open()`/`open_encrypted()` from the same scan `load_vector_indexes` already runs (no extra I/O cost beyond one more cheap `read_dir`), cleared per-pair by `create_vector_index` once a fresh index replaces the artifact.
- `resolve_vector_write_route()` replaces the old label-oblivious `set_value_is_indexed_vector_param`, wired into all three `$param` vector-write paths (`MERGE`, `MATCH...SET`, `UNWIND...MATCH...SET`). A write against a blocked pair now fails loudly, naming the label, property, exact quarantine file, and the `createVectorIndex` recovery step.
- Documented, not silently accepted: `quarantined_vector_writes` is scoped to one `GraphDb` instance, in-memory, seeded once at `open()`. A second handle on the same path (same or different process) computes its own copy; nothing reconciles them. Real limitation, out of scope here — see the doc comment on the field.

**`crates/sparrowdb-node/src/lib.rs`**
- No code change was needed for the throw itself — `execute()`/`executeWithParams()` already convert any `Result::Err` to a thrown exception, so JS callers get the loud refusal for free.
- Exposes `VectorIndexFailureKind` as `kind: "live_unserviceable" | "quarantined_no_live_index"` on `VectorIndexLoadFailure` — a plain documented string, not a new napi enum type, matching the file's existing convention and minimizing new surface.
- Corrects a doc comment that asserted the diagnostic was "the only usable signal" for #451 and that writes "are dropped" — both false as of this PR.

**`npm/sparrowdb/index.d.ts`**
- Hand-patched to match the Rust doc comment changes exactly (same `kind` field, same corrected prose). **Not regenerated** — a full napi codegen run risks pulling in unrelated drift that belongs to #513. The Rust doc comment is the actual source of truth; the next real regen will reproduce this by construction. Flagging explicitly so nobody "fixes" the `.d.ts` back to the stale prose later.

**Tests**
- New `crates/sparrowdb/tests/regression_451_quarantine_write_guard.rs`: 6 e2e tests covering all three write paths, the never-indexed unaffected case, the cross-label collision, and the create-index-clears-the-guard recovery path.
- `npm/sparrowdb/test/integration.test.js`: `assertFailureShape()` now validates `kind` per scenario; the existing `#451` reopen test gained a functional assertion that `executeWithParams` throws on the quarantined pair through the real Node binding, naming label/prop/`createVectorIndex`.

## Testing

**Rust**, `CARGO_TARGET_DIR=/Volumes/Dev/target-451`:
- `cargo build -p sparrowdb -p sparrowdb-node` — clean.
- `cargo fmt --check` — clean.
- `cargo clippy -p sparrowdb -p sparrowdb-execution -p sparrowdb-node --all-targets --keep-going` — clean.
- `cargo test -p sparrowdb --no-fail-fast` — clean, `EXIT_CODE=0`, 191 `test result: ok` blocks, 0 `FAILED` across 2404 lines of output (includes doc-tests).
- `cargo test -p sparrowdb --test regression_451_quarantine_write_guard` — 6/6 pass.

**Pre-fix confirmation** (`git stash` the 4 source files, keep the test, run, `git stash pop`) — regression tests fail against the pre-fix commit as required:
```
error must be actionable — missing `Doc`; got: invalid argument: property value is a list,
which has no scalar storage representation and cannot be written as a property value
```
```
a vector written to a label with no index of its own must not be silently swallowed
just because a DIFFERENT label happens to index a property with the same name:
QueryResult { columns: [], rows: [] }
```

**JS**, real compiled binary (`cargo build --release -p sparrowdb-node`, artifact copied to `npm/sparrowdb/sparrowdb.node`, matching CI's exact build steps):
- `npx tsc -p tsconfig.json` — clean.
- `node --test npm/sparrowdb/test/*.test.js` — **50/50 pass**, 0 fail (matches #517's count; includes my new functional #451 assertion).

## Could not verify

- **Cross-process concurrent access to `quarantined_vector_writes`** — not tested against a second `GraphDb` handle (same or different process) attached to the same db root mid-session. Documented as out of scope on the field itself.
- **napi codegen drift** — I did not run the actual napi build/codegen pipeline; the `.d.ts` change is hand-matched. If a real regen produces something different from my hand patch beyond the intended field, that's for #513 to reconcile, not this PR.
- **Windows / Linux native builds** — only verified darwin-arm64 locally, matching this machine.

## Not merging — per instructions, opening for review only.

### DIFF
--- a/crates/sparrowdb-node/src/lib.rs
+++ b/crates/sparrowdb-node/src/lib.rs
@@ -293,6 +293,7 @@ pub struct VectorIndexHealth {
 ///   prop: string;    // vector property the index was built on
 ///   path: string;    // exact file on disk; the machine-readable field
 ///   reason: string;  // human-readable; see the caveat below
+///   kind: string;    // "live_unserviceable" | "quarantined_no_live_index" — see below
 /// }
 /// ```
 ///
@@ -301,6 +302,20 @@ pub struct VectorIndexHealth {
 /// RECONSTRUCTED rather than recovered, because #442 reports the original
 /// decode failure only in the error it returns at quarantine time and writes no
 /// sidecar. Do not parse it and do not present it as the authoritative cause.
+///
+/// `kind` is the machine-checkable version of the same distinction `reason`
+/// only describes in prose (#451):
+/// - `"live_unserviceable"` — a `.bin` is present but does not resolve or does
+///   not load. `SparrowDB.open` (the Rust/other-binding equivalent) refuses to
+///   start while this is true; there is no running store to guard a write on.
+/// - `"quarantined_no_live_index"` — a `.bin.corrupt.<millis>` quarantine
+///   artifact is present and no live index currently serves the pair. The
+///   store **opened successfully** and looks healthy, but
+///   `db.executeWithParams` now throws on a vector write against this exact
+///   `(label, prop)` instead of silently discarding it — see
+///   `SparrowDB.prototype.executeWithParams`. Call `db.createVectorIndex(...)`
+///   to clear it; the entry then moves from `active` to `historical` on the
+///   next `vectorIndexDamageScan`/`vectorIndexLoadFailures` call.
 ///
 /// The `//` comments above are deliberate rather than nested doc comments: a
 /// close-comment marker inside a generated JSDoc block ends it early, which is
@@ -318,6 +333,10 @@ pub struct VectorIndexLoadFailure {
     /// Human-readable description. Reconstructed for quarantine artifacts;
     /// not authoritative and not machine-parseable.
     pub reason: String,
+    /// `"live_unserviceable"` or `"quarantined_no_live_index"` — see the
+    /// struct doc comment. Check this before deciding whether a vector write
+    /// against this `(label, prop)` would currently succeed or throw.
+    pub kind: String,
 }
 
 /// Convert the crate-level health report into its napi shape.
@@ -326,6 +345,16 @@ pub struct VectorIndexLoadFailure {
 /// valid UTF-8 must still be reported, and these functions have no error
 /// channel by design.
 fn to_damage_report(health: ::sparrowdb::VectorIndexHealth) -> VectorIndexDamageReport {
+    fn kind_str(kind: ::sparrowdb::VectorIndexFailureKind) -> String {
+        match kind {
+            ::sparrowdb::VectorIndexFailureKind::LiveUnserviceable => {
+                "live_unserviceable".to_owned()
+            }
+            ::sparrowdb::VectorIndexFailureKind::QuarantinedNoLiveIndex => {
+                "quarantined_no_live_index".to_owned()
+            }
+        }
+    }
     fn convert(v: Vec<::sparrowdb::VectorIndexFailure>) -> Vec<VectorIndexLoadFailure> {
         v.into_iter()
             .map(|f| VectorIndexLoadFailure {
@@ -333,6 +362,7 @@ fn to_damage_report(health: ::sparrowdb::VectorIndexHealth) -> VectorIndexDamage
                 prop: f.prop,
                 path: f.path.to_string_lossy().into_owned(),
                 reason: f.reason,
+                kind: kind_str(f.kind),
             })
             .collect()
     }
@@ -471,11 +501,21 @@ impl SparrowDB {
     /// - **already quarantined** — `hnsw_<label>_<prop>.bin.corrupt.<millis>`,
     ///   bytes a previous load attempt moved aside (#442).
     ///
-    /// The second arm is why this is the only usable health signal for #451:
-    /// once a damaged file has been quarantined, the *first* `open` has already
-    /// thrown and every later `open` succeeds with the index silently absent,
-    /// so vector writes for that pair are dropped and searches return nothing.
-    /// Nothing else distinguishes that store from a healthy one.
+    /// The second arm used to be the *only* usable signal for #451: once a
+    /// damaged file had been quarantined, the *first* `open` had already
+    /// thrown and every later `open` succeeded with the index silently
+    /// absent, and nothing else distinguished that store from a healthy one
+    /// — vector writes for that pair were dropped or, after #475, rejected
+    /// with a message naming neither the pair nor the quarantine.
+    ///
+    /// That is fixed as of #451: a vector write against a `(label, prop)` in
+    /// this state now throws, naming the label, the property, and this exact
+    /// quarantine artifact — see `SparrowDB.prototype.executeWithParams`.
+    /// This diagnostic remains the way to find the condition *before* a write
+    /// hits it, and each entry's `kind` field
+    /// (`VectorIndexLoadFailure.kind === "quarantined_no_live_index"`) is now
+    /// the machine-checkable way to tell it apart from a `"live_unserviceable"`
+    /// entry, which `open()` refusing to start already makes loud on its own.
     ///
     /// **Read-only since #456.** It used to quarantine the live entries it
     /// probed, so a second call answered differently from the first and the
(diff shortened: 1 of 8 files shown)


ryaker:
**Independently verified the headline finding** — the cross-label silent drop — with my own probe rather than relying on the PR's tests. It reproduces on `main` and is fixed here.

Fixture: a vector index on `(Doc, embedding)`; a `:Note` node with the same property name and no index of its own. Then `MATCH (n:Note {id:'n1'}) SET n.embedding = $emb` with a `Value::Vector` parameter.

```
main @ 17f9a1a     set_result = Ok(rows=0)      read_back = [[Null]]   -> SILENT DROP
this branch        set_result = Err(...vector has no scalar
                                storage representation...)             -> REJECTED (loud)
```

`Ok` with nothing stored and no error is real silent data loss, and it is **not** what #451 describes. Good find — it would have survived a fix that only threaded the quarantine check, and it was only reachable because routing keyed on the property name alone.

**I got this wrong twice on the way, which is worth recording** since it shaped my earlier direction to you:

1. My first probe used a list literal (`SET n.embedding = [1.0, 2.0, 3.0]`) and failed with `SET property value must be a literal or $parameter` — it never reached the vector path at all. My probe was wrong, not the code.
2. More substantively: when I briefed this work I said a vector write to a *never-indexed* property is legitimate and should be stored as an ordinary node property. **That is not this engine's behaviour** — #475 already made every non-scalar property write loud, because a vector has no scalar storage representation. So your correction to the issue's premise was more right than I credited at the time, and the post-fix `Err` above is the correct outcome for this case, not a missing feature.

**Two review notes, neither blocking:**

- `node_label`'s `unwrap_or("")` fallback would silently route a node with an unknown `label_id` to the empty-label key. I checked `catalog.list_labels()` — it maps over all entries with no filtering, so **label_id 0 is included** and the fallback can't misfire on the label-0-is-a-real-label trap that has bitten probes in this repo before. Defensive only; fine as-is.
- The post-fix error here is #475's generic message. Correct by design — this pair never had an index, so it should *not* get the quarantine-specific error — but it doesn't tell the caller that `createVectorIndex(...)` is what they want. Worth a follow-up for the never-indexed case; not for this PR.

The rest of the report checks out: doc corrected in both places, `.d.ts` hand-patched rather than regenerated (with `tsc` run against it to catch the #449 class), and the multi-process limitation documented on the field instead of half-fixed. Holding for CI.

ryaker:
Merging. Review state on the record, so the green checks aren't read as more than they are: **`claude-review` no-opped here** (3 turns, $0.124 — a real one on #514 was 10 turns/$5.90), and **CodeRabbit is rate-limited repo-wide**. No external reviewer saw this PR.

What I verified myself, in a separate worktree with its own `CARGO_TARGET_DIR`:

- **The cross-label silent drop reproduces on `main` and is fixed here** — `Ok(rows=0)` + `[[Null]]` before, a loud `Err` after. That is genuine silent data loss and it is not what #451 describes.
- **The `unwrap_or("")` label fallback is safe**: `catalog.list_labels()` maps over all entries unfiltered, so `label_id` 0 — a real label in this codebase, not a sentinel — is included and cannot silently route to the empty-label key.
- Doc corrected in **both** places, `.d.ts` hand-patched rather than regenerated with `tsc` run against it, and the multi-process limitation documented on the field rather than half-fixed.

Also recorded: I briefed this work with a wrong premise — that a vector write to a never-indexed property is stored as an ordinary node property. It isn't; #475 rejects it outright, since a vector has no scalar storage representation. The author's correction to the issue's framing was more right than I credited at the time, and the resulting `Err` is correct behaviour rather than a gap.

Issue #521, opened 2026-08-15
hop.rs has the #479 zero-sentinel bug on 4 get_node_raw sites — null-bound $param filters silently under-match on traversals


ryaker:
Found by the #472/#479 agent while proving its own fix, and **independently confirmed here.**

#479 states that the read-path call sites are safe because "`scan.rs`'s main MATCH scan, `path.rs`, `hop.rs`, `pipeline_exec.rs` — all route through `get_node_raw_nullable` + `filter_map`". **That is wrong about `hop.rs`.** Measured on `main` @ `17f9a1a`:

| file | `get_node_raw(` | `get_node_raw_nullable(` |
|---|---|---|
| **hop.rs** | **4** | **0** |
| path.rs | 0 | 0 |
| pipeline_exec.rs | 0 | 5 |
| scan.rs | 7 | 10 |
| mutation.rs | 2 | 0 |
| expr.rs | 4 | 0 |

`hop.rs` uses the plain accessor exclusively, at lines **240, 341, 602, 671**, feeding ~14 `matches_prop_filter` call sites across two-hop / n-hop / friend-of-friend traversal.

(`path.rs` calls neither, so it is safe for a different reason than #479 gives. `pipeline_exec.rs` is genuinely nullable-clean.)

## Consequence

`read_col_slot` zero-sentinels an absent slot to `Ok(0)`, which decodes to `Value::Int64(0)` and never `None`. So `stored_val.is_none()` is always false on these paths, and a **null-bound `$param` filter can never match the node whose property is genuinely absent**:

```
MATCH (a)-[:R]->(b {tag: $t})     with $t bound to null
```

silently under-matches. Fails safe (too few rows, never a wrongful write) — same class and same direction as #472, but on the traversal path rather than the pipeline `WITH … MATCH` path.

## Fix

Mechanically identical to what #472/#479's PR applies elsewhere: `get_node_raw_nullable` + `filter_map`, so absent columns arrive as `None`.

**Do not** change `read_col_slot`'s zero-sentinel to fix this — that behaviour is deliberate (SPA-406, commit `49246b3`) and changing it touches every reader in the storage layer.

## Why it was left out

Deliberately out of scope for the #472/#479 PR — separate file, larger diff, and not needed to prove that fix correct. Filed rather than folded in, which was the right call.

Note the same table shows `mutation.rs` (2) and `expr.rs` (4) using the plain accessor; those are the three sites #479 *does* name and are addressed by that PR.

[...]

PR #522, opened 2026-08-15
fix(engine): fail closed on unresolvable pipeline filters, fix absent-vs-zero conflation

### LINKED ISSUE #472: matches_prop_filter_with_binding conflates unresolved with Null (under-match sibling of #467)
Sibling of #467, **opposite direction and lower severity** — filed separately so it isn't mistaken for the data-corruption case.

`matches_prop_filter_with_binding` (`crates/sparrowdb-execution/src/engine/scan.rs:635`) — used by the pipeline `WITH … MATCH` re-traversal stage, where row bindings *are* threaded through — has the same "unresolved expression treated as `Value::Null`" conflation in its `Expr::Var` arm that #467 fixed in `matches_prop_filter_static`.

The consequence differs. Its match logic is `(None, Value::Null) => true, _ => false`, so an unresolved expression fails toward **too few** matches, never toward every node. It silently under-returns rather than silently over-writing. Not exploitable as data corruption, but it is still a query that quietly returns the wrong answer.

Confirmed by test during the #467 audit (`audit_pipeline_stage_propaccess_via_binding_does_not_widen`, scratch only — not carried into #471).

Deliberately excluded from #471: mixing an under-match fix into a data-corruption fix would have obscured the severity of both. The right fix is likely the same `is_filter_expr_resolvable` split introduced there, applied to this function.

Found during the #467 audit.

### LINKED ISSUE #479: Null-bound $param filter behaves differently on read vs mutation (get_node_raw zero-sentinels absent columns)
After #471, a pattern-property filter whose `$param` is *genuinely* bound to null matches nodes whose property is absent — on the **read** path. On the **mutation** path it matches nothing.

```
nodes: {id:'a'}  (no `tag`),  {id:'b', tag:'x'}

READ:  MATCH (n:Item {tag: $t}) RETURN n.id       $t = Null  ->  [["a"]]   correct
SET:   MATCH (n:Item {tag: $t}) SET n.val = 99    $t = Null  ->  nothing updated
```

## Cause

#471's `Value::Null => stored_val.is_none()` branch depends on absent columns arriving as `None`. That precondition holds for the read-path call sites (`scan.rs`'s main MATCH scan, `path.rs`, `hop.rs`, `pipeline_exec.rs` — all route through `get_node_raw_nullable` + `filter_map`), but not for three others that use plain `get_node_raw`:

- `mutation.rs::node_matches_prop_filter` (~line 283) — every SET/DELETE
- `expr.rs::find_node_by_props` (~line 820) — MERGE-style existence lookups
- the EXISTS-pattern check in `expr.rs` (~line 706)

`get_node_raw`'s underlying `read_col_slot` (`node_store.rs:619-637`) zero-sentinels an absent slot to `Ok(0)`, which decodes to `Value::Int64(0)` — never `None`. So on those three sites `stored_val.is_none()` is always false and a null-bound filter can never match, even for the node whose property is genuinely absent.

## Severity

Fails **safe**: a null-bound filter matches nothing rather than everything, so there is no wrongful mutation. It is nonetheless wrong — the inline equivalent of `SET … WHERE prop IS NULL` silently no-ops instead of updating the intended node, with no error.

Note this does not weaken #467's fix: `is_filter_expr_resolvable` runs *before* the `Value::Null` arm, so a missing or unbound filter is still blocked on mutations. This affects only the narrower explicitly-null-bound `$param` semantic that #471 introduced.

## Unverified

Whether `find_node_by_props`'s participation makes a null-matched MERGE create a duplicate node. Reasoned plausible, not reproduced — worth checking before treating as more than a maybe.

## Fix direction

Make the three mutation-side call sites source props the same way the read sites do (`get_node_raw_nullable`), so "absent" and "zero" stop being the same value. That distinction is the root cause of this and of #473/#475, where a null *write* becomes `Int64(0)` — absent-vs-zero conflation is worth fixing once, structurally, rather than three times.

Found by adversarial review of #471.

ryaker:
closes #472, closes #479

## Root cause — #472

`matches_prop_filter_with_binding` (`engine/scan.rs`, the pipeline `WITH … MATCH` re-traversal stage) treated an unresolved `Expr::Var` — a bare variable not present in the row `binding` (a typo, or a correlated/outer variable never carried through from the preceding `WITH`) — as `Value::Null` by default (`binding.get(v).cloned().unwrap_or(Value::Null)`). Combined with the `(None, Value::Null) => true` match arm, this silently matched every node whose own property happened to be absent — a symptom-free wrong answer, not the fail-open "match everything" #467 guarded against, but wrong all the same. Fix: apply the same `is_filter_expr_resolvable_scoped` gate #467/#471 introduced for the static filter path, extended so a variable present in `binding` counts as locally resolvable (mirrors the existing `locals` mechanism for comprehension-bound loop variables). An unresolvable filter now fails closed (the node does not match) instead of defaulting to null.

## Root cause — #479

Three mutation-side call sites read node props via `get_node_raw`, whose `read_col_slot` zero-sentinels an absent column to `Ok(0)` — indistinguishable from a stored `Int64(0)`. Every read-path call site instead uses `get_node_raw_nullable` + `filter_map` to drop absent columns entirely, so `matches_prop_filter_static`'s `Value::Null => stored_val.is_none()` arm works correctly there. On the three non-nullable sites it never fired, so a genuinely null-bound `$param` filter matched on read but nothing on SET/DELETE/EXISTS. Fixed by routing all three through the same nullable-accessor + drop-absent pattern already established on every read path, per repo guidance to avoid touching `read_col_slot`'s zero-sentinel itself (SPA-406, commit `49246b3` — changing it would alter every reader in the storage layer, a blast radius far beyond these two issues).

## What changed, per file

- **`crates/sparrowdb-execution/src/engine/scan.rs`**
  - `matches_prop_filter_with_binding`: added the resolvability gate (#472).
  - `collect_pipeline_match_rows` and the three `get_node_raw` calls inside `execute_pipeline_match_stage` (leading-match scan, src-candidate scan, dst-node scan in the hop loop): switched to `get_node_raw_nullable` + `filter_map`. **This was not optional** — while building an end-to-end test for #472's "genuinely null-bound variable must still match the absent-prop node" control case, the test failed even with the resolvability gate in place, because these four sites never produced `stored_val = None` regardless of the gate. Fixing #472 correctly required fixing this #479-shaped bug in the same file; isolated to confirm both fixes carry independent weight (see Verification).
- **`crates/sparrowdb-execution/src/engine/mutation.rs`** — `node_matches_prop_filter`: routed through `get_node_raw_nullable` + `filter_map`. Reaches every SET, DELETE (node), DELETE (edge, via `scan_match_mutate_edges`), and MATCH…CREATE / MATCH…MERGE-rel candidate scan (all route through `scan_nodes_for_label_with_index`'s fallback path).
- **`crates/sparrowdb-execution/src/engine/expr.rs`** — `find_node_by_props` (shortestPath endpoint resolution by label+props — *not* MERGE, see below) and the EXISTS-pattern dst-node check in `eval_exists_subquery`: same fix.
- **`crates/sparrowdb/tests/regression_472_479.rs`** — 9 new tests (added).

## Test output — quoted, pre-fix

All source changes stashed, tests rerun against unfixed `main`:

```
delete_with_null_bound_param_removes_only_absent_prop_node ... FAILED
  got [[String("a")], [String("b")]]     expected [[String("b")]]
set_with_null_bound_param_updates_only_absent_prop_node ... FAILED
  got [[a,1],[b,1]]                      expected [[a,99],[b,1]]
exists_subquery_dst_prop_filter_matches_absent_prop_node ... FAILED
  got []                                 expected [[a1]]
genuinely_null_bound_binding_var_matches_only_absent_prop_node ... FAILED
  got []                                 expected [[a]]
```

The other 5 tests (read control, resolvable-var controls, unresolved-var test) passed pre-fix too — expected, since they either don't depend on the fix or are guarded by the zero-sentinel bug masking the #472 symptom (see next section).

**Isolating the #472 gate specifically**: with the zero-sentinel fix applied but `is_filter_expr_resolvable_scoped` disabled (`if false && !...`), `unresolved_var_in_pipeline_match_prop_filter_matches_nothing` failed: `got [[String("a")]]  expected []`. This confirms the resolvability gate is independently load-bearing, not just riding on the nullable-read fix — two fixes landed together, and each was shown to matter on its own.

## Blast radius checked

- `node_matches_prop_filter`'s callers: `scan_match_mutate` (SET/DELETE), `scan_match_mutate_edges` (edge DELETE), `scan_nodes_for_label_with_index` (MATCH…CREATE, MATCH…MERGE-rel candidate scans). All changes are correctness-only for the narrow null-bound-filter case; behavior is unchanged for every other filter shape.
- `find_node_by_props` is used only by `eval_shortest_path_expr` (shortestPath endpoint resolution), never by MERGE.
- Ran `regression_467`, `spa_134_multi_clause`, `regression_444_430_null_projection`, `spa_137_exists_subquery` post-rebase — all green, no regressions in the pipeline/WITH/EXISTS suites this PR touches.
- `cargo fmt` (no-op), `cargo clippy -p sparrowdb -p sparrowdb-execution --all-targets --keep-going` (clean).
- Local `cargo test --workspace --exclude sparrowdb-ruby --no-fail-fast` was still running at push time (the suite is known-slow, 60+ min); not used as a merge gate — CI runs the authoritative version of this suite plus the Ruby job this machine can't compile.

## Resolving the issue's own "Unverified" section (#479)

> Whether `find_node_by_props`'s participation makes a null-matched MERGE create a duplicate node.

Checked directly: `find_node_by_props` is **not** in MERGE's path at all — MERGE uses its own comparison logic in `write_tx.rs::merge_node` (line ~303), which has the *same* absent-vs-zero conflation in principle. But it's unreachable in practice: a null-valued MERGE property is rejected upfront ("property value is null; use a concrete value" — `helpers.rs:72/110/230`, `engine/mod.rs:1615`) before that comparison is ever reached. Confirmed by probe (`MERGE (n:Item {tag: $t})` with `$t = Null` returns `InvalidArgument`, not a duplicate). So: real bug in principle in `merge_node`, unreachable today for a different reason than guessed — worth stating explicitly so nobody re-derives it, and so it's on record if the upfront null rejection ever moves.

## Found, not fixed — filed separately

**#521**: `hop.rs` (two-hop/n-hop/friend-of-friend traversal) has the identical zero-sentinel bug on 4 `get_node_raw` call sites (lines 240, 341, 602, 671), feeding ~14 `matches_prop_filter` call sites. This contradicts #479's own claim that "path.rs, hop.rs, pipeline_exec.rs all route through `get_node_raw_nullable`" — confirmed by direct count, `hop.rs` uses the plain accessor exclusively. Not fixed here: separate file, larger mechanical diff, not required to prove this PR's fix. `path.rs` is safe for a different reason than #479 states (it calls neither accessor — no prop-filter reads at all in that file); `pipeline_exec.rs` is genuinely nullable-clean.

## Could not verify

- Full `cargo test --workspace --exclude sparrowdb-ruby --no-fail-fast` completion post-rebase — the suite takes 60+ min on this box and was still running when this PR was pushed. Will report any failures it surfaces as a follow-up; not treating it as a merge gate since CI runs the authoritative version (plus the Ruby job this machine can't compile at all).
- Whether any other write-path (beyond the two crates touched here) shares this conflation — I checked `operators.rs` (a separate columnar/vectorized scan operator) and it reads raw column values without a null-vs-zero filter comparison, so it's out of the "prop filter" family this PR is about, but I did not audit it exhaustively.
- `write_tx.rs::merge_node`'s own latent absent-vs-zero conflation (see above) is unreachable today given the upfront null-value rejection; I did not verify whether any other code path could construct a `Value::Null` `want_val` that bypasses that rejection.

### DIFF
--- a/crates/sparrowdb-execution/src/engine/expr.rs
+++ b/crates/sparrowdb-execution/src/engine/expr.rs
@@ -957,9 +957,19 @@ impl Engine {
                         .iter()
                         .map(|p| prop_name_to_col_id(&p.key))
                         .collect();
-                    match self.snapshot.store.get_node_raw(probe_id, &col_ids) {
-                        Ok(props) => {
+                    // #479: nullable accessor + drop-absent, matching every
+                    // read-path prop-filter site — see node_matches_prop_filter.
+                    match self
+                        .snapshot
+                        .store
+                        .get_node_raw_nullable(probe_id, &col_ids)
+                    {
+                        Ok(raw_props) => {
                             let params = self.dollar_params();
+                            let props: Vec<(u32, u64)> = raw_props
+                                .into_iter()
+                                .filter_map(|(c, opt)| opt.map(|v| (c, v)))
+                                .collect();
                             if !matches_prop_filter_static(
                                 &props,
                                 &dst_pat.props,
@@ -1073,8 +1083,16 @@ impl Engine {
         let params = self.dollar_params();
         for slot in 0..hwm {
             let node_id = NodeId(((label_id as u64) << 32) | slot);
-            if let Ok(raw_props) = self.snapshot.store.get_node_raw(node_id, &col_ids) {
-                if matches_prop_filter_static(&raw_props, props, &params, &self.snapshot.store) {
+            // #479: route through the nullable accessor (as the read-path scan
+            // does) and drop absent columns, so a genuinely null-bound filter
+            // can match the node whose property is actually absent instead of
+            // silently never matching via `get_node_raw`'s zero-sentinel.
+            if let Ok(raw_props) = self.snapshot.store.get_node_raw_nullable(node_id, &col_ids) {
+                let stored: Vec<(u32, u64)> = raw_props
+                    .into_iter()
+                    .filter_map(|(c, opt)| opt.map(|v| (c, v)))
+                    .collect();
+                if matches_prop_filter_static(&stored, props, &params, &self.snapshot.store) {
                     return Some(slot);
                 }
             }
--- a/crates/sparrowdb-execution/src/engine/mutation.rs
+++ b/crates/sparrowdb-execution/src/engine/mutation.rs
@@ -279,13 +279,31 @@ impl Engine {
         if props.is_empty() {
             return true;
         }
-        match self.snapshot.store.get_node_raw(node_id, filter_col_ids) {
-            Ok(raw_props) => matches_prop_filter_static(
-                &raw_props,
-                props,
-                &self.dollar_params(),
-                &self.snapshot.store,
-            ),
+        // #479: use the nullable accessor and drop absent columns from the
+        // list entirely, mirroring every read-path call site. `get_node_raw`'s
+        // `read_col_slot` zero-sentinels an absent slot to `Ok(0)` (decodes to
+        // `Value::Int64(0)`), which made `matches_prop_filter_static`'s
+        // `Value::Null => stored_val.is_none()` arm unreachable here — a
+        // genuinely null-bound `$param` filter could never match the node
+        // whose property is actually absent, even though the read path
+        // matches it correctly.
+        match self
+            .snapshot
+            .store
+            .get_node_raw_nullable(node_id, filter_col_ids)
+        {
+            Ok(raw_props) => {
+                let stored: Vec<(u32, u64)> = raw_props
+                    .into_iter()
+                    .filter_map(|(c, opt)| opt.map(|v| (c, v)))
+                    .collect();
+                matches_prop_filter_static(
+                    &stored,
+                    props,
+                    &self.dollar_params(),
+                    &self.snapshot.store,
+                )
+            }
             Err(_) => false,
         }
     }
--- a/crates/sparrowdb-execution/src/engine/scan.rs
+++ b/crates/sparrowdb-execution/src/engine/scan.rs
@@ -304,10 +304,19 @@ impl Engine {
             if self.is_node_tombstoned(node_id) {
                 continue;
             }
-            let props = match self.snapshot.store.get_node_raw(node_id, &col_ids) {
+            // Nullable read so absent columns are omitted rather than
+            // zero-sentineled, matching the main MATCH scan (#479-adjacent:
+            // this pipeline leading-match scan had the same absent-vs-zero
+            // conflation as the three sites #479 named explicitly).
+            let nullable_props = match self.snapshot.store.get_node_raw_nullable(node_id, &col_ids)
+            {
                 Ok(p) => p,
                 Err(_) => continue,
             };
+            let props: Vec<(u32, u64)> = nullable_props
+                .into_iter()
+                .filter_map(|(c, opt)| opt.map(|v| (c, v)))
+                .collect();
             if !self.matches_prop_filter(&props, &node.props) {
                 continue;
             }
@@ -377,10 +386,19 @@ impl Engine {
             if self.is_node_tombstoned(node_id) {
                 continue;
             }
-            let props = match self.snapshot.store.get_node_raw(node_id, &col_ids) {
+            // Nullable read so absent columns are omitted rather than
+            // zero-sentineled — same #479-adjacent fix as collect_pipeline_match_rows
+            // above, needed so a genuinely null-bound binding var matches the
+            // node whose property is actually absent (see matches_prop_filter_with_binding).
+            let nullable_props = match self.snapshot.store.get_node_raw_nullable(node_id, &col_ids)
+            {
                 Ok(p) => p,
                 Err(_) => continue,
             };
+            let props: Vec<(u32, u64)> = nullable_props
+                .into_iter()
+                .filter_map(|(c, opt)| opt.map(|v| (c, v)))
+                .collect();
 
             // Evaluate inline prop filters, resolving variable references from binding.
             if !self.matches_prop_filter_with_binding(&props, &node.props, binding, &params) {
@@ -507,7 +525,16 @@ impl Engine {
                 if self.is_node_tombstoned(node_id) {
                     continue;
                 }
-                if let Ok(props) = self.snapshot.store.get_node_raw(node_id, &src_col_ids) {
+                // Nullable read — see the analogous fix a few lines above.
+                if let Ok(nullable_props) = self
+                    .snapshot
+                    .store
+                    .get_node_raw_nullable(node_id, &src_col_ids)
+                {
+                    let props: Vec<(u32, u64)> = nullable_props
+                        .into_iter()
+                        .filter_map(|(c, opt)| opt.map(|v| (c, v)))
+                        .collect();
                     if self.matches_prop_filter_with_binding(
                         &props,
                         &src_pat.props,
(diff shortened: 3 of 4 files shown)


ryaker:
**Verified independently** with my own probe against both sides, rather than relying on the PR's tests.

Fixture: two `:Item` nodes — `{id:'a'}` with **no** `tag` property, and `{id:'b', tag:'x'}`. Both start at `val = 1`. Filter is `{tag: $t}` with `$t` genuinely bound to `Value::Null`.

```
                    main @ c8ed523                        this branch
read   MATCH ... RETURN n.id      [["a"]]                 [["a"]]        (unchanged)
set    MATCH ... SET n.val = 99   [[a,1],[b,1]]  NO-OP    [[a,99],[b,1]]  fixed
delete MATCH ... DELETE n         [[a],[b]]      NO-OP    [[b]]           fixed
```

So on `main` the read path matches the absent-property node and **both mutation paths silently match nothing** while returning `Ok` — confirming #479 exactly, and showing it extends to `DELETE`, not just `SET`.

**The check I care about most: `b` is still not matched.** A fix in this area can easily overshoot from under-matching into *over*-matching, which is #467's direction — where an unresolved filter made `SET` write **every** node. It didn't: `b` (which genuinely has `tag:'x'`) is untouched by both the SET and the DELETE. Fail-closed stayed fail-closed for the resolvable case and started matching correctly for the genuinely-null case, which is the whole point.

**Two things in this PR I want to credit specifically**, because both are the difference between a fix and a demonstrated fix:

1. **Isolating the gate.** Disabling only `is_filter_expr_resolvable_scoped` while keeping the nullable-read fix, and showing a test still fails, proves the gate is load-bearing rather than decorative. Two changes landing together very often means one is doing nothing; that's rarely checked.
2. **Fixing the 4 extra `scan.rs` zero-sentinel sites was not scope creep.** Without them `stored_val` was never `None`, so #472's gate was untestable end-to-end. Required to prove the fix, same file, correct call.

The MERGE resolution is stated the right way too — *real bug in principle, unreachable in practice, for a different reason than the issue guessed* — rather than a bare "MERGE is fine". That survives someone later moving the upfront null rejection.

`hop.rs` exposure is tracked as **#521**; I confirmed those counts independently (`hop.rs` 4 raw / 0 nullable, `pipeline_exec.rs` 0 / 5, `path.rs` 0 / 0) and the issue records that #479's own text is wrong about `hop.rs` and imprecise about `path.rs`.

Holding for CI.

ryaker:
Merging — CI fully green including benchmarks.

**Review state on the record:** `claude-review` no-opped here (5 turns, $0.176; a real run on #514 was 10 turns/$5.90) and CodeRabbit is rate-limited repo-wide. **No external reviewer saw this PR** — that is now three tonight, so the green checkmarks represent CI plus my own verification, nothing more.

What I verified independently, in a separate worktree, before merging:

- **#479's read/mutation divergence reproduces on `main` and is fixed here** — and it extends to `DELETE`, not just `SET`. Both mutations returned `Ok` while doing nothing.
- **The fix did not overshoot into fail-open.** The node that genuinely has `tag:'x'` is still untouched by both the SET and the DELETE. That was the check worth running: the neighbouring bug #467 was the over-matching direction, where an unresolved filter made `SET` write every node, and under-match/over-match are one edit apart.
- **#521's counts**, which this PR correctly leaves alone: `hop.rs` 4 raw / 0 nullable, `pipeline_exec.rs` 0 / 5, `path.rs` 0 / 0.

Two things in this PR are better practice than the fix itself and worth repeating: isolating the resolvability gate to prove it is load-bearing independently of the nullable-read fix, and stating the MERGE resolution as *real bug in principle, unreachable in practice, for a different reason than the issue guessed* rather than a bare "MERGE is fine".

PR #526, opened 2026-08-16
fix(execution): route hop.rs's 6 zero-sentineled read sites through nullable accessors

### LINKED ISSUE #521: hop.rs has the #479 zero-sentinel bug on 4 get_node_raw sites — null-bound $param filters silently under-match on traversals

ryaker:
## Summary

`read_col_slot` (storage layer) zero-sentinels an absent column to `Ok(0)`, which decodes to `Value::Int64(0)` and never `None`. This defect exists on **two independent call paths in `hop.rs`**, not just the one the issue named:

1. **4 `get_node_raw` sites** in `execute_one_hop` — the individual-node-read accessor. This is what issue #521 named (its own survey grepped for `get_node_raw(` occurrences).
2. **2 `batch_read_node_props` sites** — the *batch*-read accessor, used by `execute_one_hop`'s dst-side and `execute_two_hop`'s friend-of-friend side. **Not in the original issue text** — its grep-based survey only counted `get_node_raw(` and missed this differently-named function entirely, the same class of gap #521 itself calls out in #479's survey ("#479 states that the read-path call sites are safe... That is wrong about hop.rs"). I found this mid-task, flagged it to @ryaker before proceeding, and was told to expand scope to fix it in this PR rather than defer it — see the corrected picture below.

Both families are fixed here. `read_col_slot` and `batch_read_node_props`/`get_node_raw` themselves are **unchanged** — every other reader in the storage layer is unaffected. Only the 6 call sites inside `hop.rs` were touched.

closes #521

## The corrected picture (my issue's survey was wrong)

Family 2 turned out to matter *more* in practice than family 1: `get_node_raw` at the individual-read sites is mostly a rare/unreachable fallback (see below), but `batch_read_node_props` is the **primary** read path for two prop filters — the common case, not an edge case. One of its two sites additionally carried a manual `.filter(|&(_, v)| v != 0)` workaround that happened to fix the null-under-match symptom by accident, while introducing the *opposite* bug: a genuinely stored `Int64(0)` was silently misread as absent — live wrong data on `main` for any integer property that happens to be zero, not just a null-filter edge case.

## What changed (1 file, `crates/sparrowdb-execution/src/engine/hop.rs`)

All 6 sites converted to the existing nullable read paths already used elsewhere in this file (`execute_two_hop`/`execute_n_hop`'s other reads):

| Site | Accessor | Role | Query shape that reaches it |
|---|---|---|---|
| `execute_one_hop` src_props | `get_node_raw` → `read_node_props` | forward pass, individual read | `MATCH (a:A {tag:$t})-[:R]->(b:B) ...` — directed, source-side filter |
| `execute_one_hop` dst_props (fallback) | `get_node_raw` → `read_node_props` | batch-miss fallback, individual read | **Not independently reachable** — see "Could not verify" |
| `execute_one_hop` dst_props (**primary**) | `batch_read_node_props` → `batch_read_node_props_nullable` | forward pass, batch read | Any one-hop query with a dst-side filter and ≥1 neighbor found via CSR/delta (the common case) |
| `execute_one_hop` b_props | `get_node_raw` → `read_node_props` | backward "a"-role, individual read | Undirected: `MATCH (a:P {tag:$t})-[r]-(b:P) ...`, only when `dir == EdgeDir::Both` |
| `execute_one_hop` a_props | `get_node_raw` → `read_node_props` | backward "b"-role, individual read | Undirected: `MATCH (a:P)-[r]-(b:P {tag:$t}) ...`, only when `dir == EdgeDir::Both` |
| `execute_two_hop` fof_props (**primary**) | `batch_read_node_props` + `.filter(v != 0)` → `batch_read_node_props_nullable` | forward-forward branch, batch read | `MATCH (a)-[:R]->(m)-[:R]->(f) ...` with an `f`-side filter, second hop Outgoing |

`read_node_props` (individual sites) and the new `batch_read_node_props_nullable` calls (batch sites) both drop absent columns from the props vec entirely (`filter_map`), rather than zero-filling — matching the pattern the rest of this file already uses for its correct sites.

**Loop-nesting check**: all 6 conversions are 1:1 substitutions at the exact call site — same position in the same loop, same arguments (plus the `filter_map`/`Option` unwrap where the batch signature changed), no code moved between loop levels. The two batch calls remain computed once per `src_slot` (outside their respective inner per-neighbor loops), exactly as before — I checked this specifically given the repo's prior 44x HNSW regression from a block landing in the wrong loop during a port.

## Tests (`crates/sparrowdb/tests/regression_521_hop_absent.rs`)

5 tests total, each isolating one site (4 covered end-to-end; the 5th site is the unreachable fallback, documented not faked):

1. **`src_prop_filter_forward_hop_matches_absent_node`** (individual, forward src) — directed one-hop. Pre-fix: `got []` (confirmed against `44f27de~1`).
2. **`undirected_backward_a_role_filter_matches_absent_node`** (individual, backward a-role) — undirected. Pre-fix: `got []` (confirmed against `44f27de~1`).
3. **`undirected_backward_b_role_filter_matches_absent_node`** (individual, backward b-role) — undirected. **Its expected value changed with the scope expansion**: fixing the dst-batch path (site 3 in the table) means a forward-pass row that used to be excluded only *because of* the batch bug now correctly appears. Quoted outputs:
   - Against `44f27de~1` (pre any #521 fix): `got []`
   - Against the commit with only the 4-site individual-read fix (`f014bf6`, i.e. before batch expansion): `got [[Bob, Alice]]` — one row, not the current expected two
   - Against this PR's full fix: `[[Bob,Alice],[Carol,Alice]]` ✓ (the currently-expected value)
4. **`one_hop_dst_batch_filter_matches_absent_and_preserves_real_zero`** (**batch**, dst primary path) — 4 neighbors off one src node, forcing `unique_dst_slots.len() == 4` through the actual batch read, not the fallback. Two assertions:
   - Null filter must match only the 2 genuinely-absent nodes. Pre-fix (`f014bf6`): `got []` — under-match, every neighbor's tag column reads back as `Some(0)` regardless of absence.
   - A literal-`0` filter must find only the genuinely-stored-0 node. Pre-fix (`f014bf6`, isolated via a throwaway probe since the first assertion already fails and short-circuits the function): `got [f_absent, f_absent2, f_zero]` — over-match, absent columns are indistinguishable from a real `Int64(0)` at the raw-byte level (both are `0u64`) without the null-bitmap check.
5. **`two_hop_fof_batch_filter_matches_absent_and_preserves_real_zero`** (**batch**, fof primary path, forward-forward branch) — same shape one hop further out, forces the batch through `all_fof_slots.len() == 4`. Two assertions:
   - Null filter pre-fix (`f014bf6`): `got [f_absent, f_absent2, f_zero]` — the old `.filter(v != 0)` workaround over-matches, since it also treats real zero as absent.
   - Literal-`0` filter pre-fix (isolated via probe): `got []` — the workaround drops `f_zero` from the props vec entirely (raw value `0`), so it can never be found again, even by an exact-zero filter.

**Over-match direction verified in every test**: each fixture includes a node with the property genuinely present (non-zero string, or non-empty other-tag node), and each assertion is an exact-set match.

**Could not verify** — read this section first:
- `execute_one_hop`'s dst-props individual-read fallback is **unreachable through the public query API as currently structured**: `unique_dst_slots` (dedup of `all_neighbors`) and the outer per-neighbor loop iterate the same collection, `dst_slot_to_idx` indexes every entry of `unique_dst_slots`, and both `batch_read_node_props` and `batch_read_node_props_nullable` always return exactly one row per requested slot. `dst_slot_to_idx.get(&dst_slot)` can never miss. I converted it anyway (consistency, and in case a future change — e.g. a partial/streamed batch — makes it reachable), but it's unverified by test.
- I did **not** run `cargo test --workspace` locally (excludes `sparrowdb-ruby`, which doesn't compile on this machine; redundant with CI per repo `CLAUDE.md`).
- I have **not** verified interaction with `execute_variable_length`/varlen-hop paths — those don't call into the functions I touched per the `scan.rs` dispatch logic I read, so I believe they're unaffected, but didn't add tests for them specifically.
- The two `zero_result` pre-fix outputs quoted above were captured via a throwaway probe test (not committed) rather than the shipped test itself, because each shipped test's first assertion (`null_result`) already fails and panics before the second assertion (`zero_result`) would run — so the shipped test file only directly demonstrates one failure mode per test pre-fix, with the other confirmed separately. Both probe runs were against the same `f014bf6` checkout as the corresponding shipped-test failures, immediately before restoring the fix.

## Verification run (all local, `CARGO_TARGET_DIR=/Volumes/Dev/target-521`)

- `cargo fmt` — clean
- `cargo clippy -p sparrowdb -p sparrowdb-execution --all-targets --keep-going` — 0 warnings
- `cargo check -p sparrowdb` — passes (merge-discipline gate)
- `cargo test -p sparrowdb-execution --lib` — 50/50 pass
- Targeted e2e suites (`--no-fail-fast`), all pass: `regression_521_hop_absent` (new, 5), `regression_467` (11), `regression_472_479` (9), `regression_487_488_edge_direction` (10), `spa_193_undirected_pattern` (3), `spa_201_csr_backward` (5), `spa_299_phase4_parity` (17), `spa_188_two_hop_where` (4), `spa_200_batch_hop_perf` (5), `spa_241_multihop_props` (4), `spa_252_three_hop_binding` (4), `spa_263_two_hop_agg` (3), `spa_263_two_hop_null` (3)
- Did not run `cargo test --workspace` — see "Could not verify"
- Every new test independently confirmed to fail against the relevant pre-fix commit with the quoted output above, then confirmed to pass again after restoring the fix.

### DIFF
--- a/crates/sparrowdb-execution/src/engine/hop.rs
+++ b/crates/sparrowdb-execution/src/engine/hop.rs
@@ -237,7 +237,14 @@ impl Engine {
                         }
                         v
                     };
-                    self.snapshot.store.get_node_raw(src_node, &all_needed)?
+                    // #521: nullable accessor + drop-absent (matches the
+                    // read_node_props helper already used by execute_two_hop /
+                    // execute_n_hop), so a genuinely null-bound $param filter
+                    // can match a node whose property is actually absent
+                    // instead of silently never matching via get_node_raw's
+                    // zero-sentinel (read_col_slot returns Ok(0) for an absent
+                    // slot, indistinguishable from a stored Int64(0)).
+                    read_node_props(&self.snapshot.store, src_node, &all_needed)?
                 } else {
                     vec![]
                 };
@@ -336,9 +343,10 @@ impl Engine {
                                 .collect()
                         } else {
                             // Fallback: individual read (e.g. delta-only slot).
-                            self.snapshot
-                                .store
-                                .get_node_raw(dst_node, &all_needed_dst)?
+                            // #521: nullable accessor + drop-absent — see the
+                            // src_props comment above for why get_node_raw's
+                            // zero-sentinel is wrong here.
+                            read_node_props(&self.snapshot.store, dst_node, &all_needed_dst)?
                         }
                     } else {
                         vec![]
@@ -599,7 +607,9 @@ impl Engine {
                             }
                             v
                         };
-                        self.snapshot.store.get_node_raw(b_node, &all_needed)?
+                        // #521: nullable accessor + drop-absent — see the
+                        // src_props comment in the forward pass above.
+                        read_node_props(&self.snapshot.store, b_node, &all_needed)?
                     } else {
                         vec![]
                     };
@@ -668,7 +678,9 @@ impl Engine {
                                 }
                                 v
                             };
-                            self.snapshot.store.get_node_raw(a_node, &all_needed)?
+                            // #521: nullable accessor + drop-absent — see the
+                            // src_props comment in the forward pass above.
+                            read_node_props(&self.snapshot.store, a_node, &all_needed)?
                         } else {
                             vec![]
                         };
--- a/crates/sparrowdb-execution/src/engine/hop.rs
+++ b/crates/sparrowdb-execution/src/engine/hop.rs
@@ -304,8 +304,19 @@ impl Engine {
 
                 // Batch-read: one fs::read() per column for all neighbors.
                 // dst_batch[i] = raw column values for unique_dst_slots[i].
-                let dst_batch: Vec<Vec<u64>> = if !all_needed_dst.is_empty() {
-                    self.snapshot.store.batch_read_node_props(
+                //
+                // #521 (broader finding, expanded scope): use the nullable
+                // variant so absent columns arrive as `None` instead of the
+                // old raw-0-as-absent sentinel `batch_read_node_props` used —
+                // same defect as `get_node_raw`'s zero-sentinel (see the
+                // src_props comment above), but this is the *primary* path
+                // for dst_props below, not a rare fallback: every neighbor
+                // slot found via CSR/delta lands in `dst_slot_to_idx` and is
+                // read from this batch; the individual `read_node_props`
+                // fallback a few lines down is the one that's effectively
+                // unreachable.
+                let dst_batch: Vec<Vec<Option<u64>>> = if !all_needed_dst.is_empty() {
+                    self.snapshot.store.batch_read_node_props_nullable(
                         effective_dst_label_id,
                         &unique_dst_slots,
                         &all_needed_dst,
@@ -340,6 +351,7 @@ impl Engine {
                                 .iter()
                                 .copied()
                                 .zip(dst_batch[idx].iter().copied())
+                                .filter_map(|(c, opt)| opt.map(|v| (c, v)))
                                 .collect()
                         } else {
                             // Fallback: individual read (e.g. delta-only slot).
@@ -1641,8 +1653,18 @@ impl Engine {
             };
 
             // Batch-read fof properties once per src_slot.
-            let fof_batch: Vec<Vec<u64>> = if !col_ids_fof.is_empty() {
-                self.snapshot.store.batch_read_node_props(
+            //
+            // #521 (broader finding, expanded scope): use the nullable
+            // variant. The old `batch_read_node_props` + `.filter(v != 0)`
+            // combination below was a *different*, worse-direction bug than
+            // #521's null-under-match: it discarded any column whose raw
+            // value was literally 0, misreading a genuinely stored
+            // `Int64(0)` as absent — silent wrong data, not just a null-filter
+            // edge case. `batch_read_node_props_nullable` distinguishes the
+            // two via the null-bitmap sidecar, so absence and stored-zero no
+            // longer collide.
+            let fof_batch: Vec<Vec<Option<u64>>> = if !col_ids_fof.is_empty() {
+                self.snapshot.store.batch_read_node_props_nullable(
                     fof_label_id,
                     &all_fof_slots,
                     &col_ids_fof,
@@ -1679,7 +1701,7 @@ impl Engine {
                                 .iter()
                                 .copied()
                                 .zip(fof_batch[idx].iter().copied())
-                                .filter(|&(_, v)| v != 0)
+                                .filter_map(|(c, opt)| opt.map(|v| (c, v)))
                                 .collect()
                         } else {
                             // Fallback: individual read (delta-only slot not in batch).
--- a/crates/sparrowdb/tests/regression_521_hop_absent.rs
+++ b/crates/sparrowdb/tests/regression_521_hop_absent.rs
@@ -1,31 +1,51 @@
-//! Regression guards for issue #521 — the same "absent conflated with zero"
-//! defect #479/#522 fixed in mutation.rs/expr.rs, found on 4 remaining
-//! call sites in hop.rs's `execute_one_hop`: `get_node_raw` zero-sentinels
-//! an absent column to `Ok(0)` (decoding to `Value::Int64(0)`, never
-//! `None`), so a genuinely null-bound `$param` prop filter could never
-//! match the node whose property is actually absent on these traversal
-//! paths — silent under-matching (same direction as #472/#479, on the
-//! traversal path rather than the pipeline/mutation path).
+//! Regression guards for issue #521 — the "absent conflated with zero"
+//! defect #479/#522 fixed in mutation.rs/expr.rs, found on two families of
+//! call sites in hop.rs:
+//!
+//! 1. **4 `get_node_raw` sites** in `execute_one_hop` (the individual-read
+//!    accessor): `read_col_slot` zero-sentinels an absent column to `Ok(0)`
+//!    (decoding to `Value::Int64(0)`, never `None`), so a genuinely
+//!    null-bound `$param` prop filter could never match the node whose
+//!    property is actually absent — silent under-matching (same direction
+//!    as #472/#479, on the traversal path rather than pipeline/mutation).
+//! 2. **2 `batch_read_node_props` sites** (the batch-read accessor used by
+//!    `execute_one_hop`'s dst-side and `execute_two_hop`'s friend-of-friend
+//!    side) — found while building this PR, not in the original issue text
+//!    (which only grepped for `get_node_raw(` and missed the differently
+//!    named batch function entirely). These are the *primary* read path for
+//!    those two prop filters, not a rare fallback, so they matter more in
+//!    practice than family 1. One of the two sites additionally carried a
+//!    manual `.filter(|&(_, v)| v != 0)` workaround that fixed the null case
+//!    by accident while breaking the opposite case: a genuinely stored
+//!    `Int64(0)` was misread as absent. Both sites now use
+//!    `batch_read_node_props_nullable`, which distinguishes absence from
+//!    stored-zero via the null-bitmap sidecar.
 //!
 //! All expected values below are derived by hand from each fixture, never
 //! captured from a prior run (repo rule — see CLAUDE.md). Each test's
 //! pre-fix failure output is quoted in the PR description, captured by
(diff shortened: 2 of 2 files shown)


ryaker:
**Holding this PR: it does not yet fix #521's headline case.** A scope decision I sent the author never reached their inbox, so this is not an oversight on their part — recording the reasoning here.

The PR's own findings section is what surfaced it. It reports that the 4th site — the `dst_props` **batch-miss fallback** — is *provably unreachable*, because `dst_slot_to_idx` always covers every `dst_slot`. That is the tell. Verified on this branch:

```
~309  batch_read_node_props(          <- non-nullable, still zero-sentinels
~338  .zip(dst_batch[idx]...)         <- the REACHABLE dst path, UNFIXED
~349  read_node_props(...)            <- fixed, but on the branch proved dead
```

If the fallback never runs, then **every reachable destination-side property read comes from the non-nullable batch.** So:

```
MATCH (a)-[:R]->(b {tag: $t})     $t bound to null
```

— the exact query shape in this issue's title — **still under-matches after this PR.** Three reachable sites are genuinely fixed (source-side, and both undirected sides), plus one dead one.

**A PR that closes #521 while its titular symptom still reproduces is worse than no PR**, because the issue stops being a signal that anything is wrong.

## Expanding scope rather than splitting

`batch_read_node_props_nullable` (`node_store.rs:1500`) already exists and is doc-commented as *"the correct API… replaces the raw-0-as-absent semantic"*. Swapping the call site is applying a decision already made, not making a new one.

Also folding in **hop.rs:1670**'s `.filter(|&(_, v)| v != 0)` in `execute_n_hop`, which is arguably worse than the bug this issue was filed for: it discards a **genuinely stored `Int64(0)`** as absent, so ordinary integer properties read back wrong in n-hop traversal with no parameters involved at all. It gets its own findings entry rather than being folded into the #521 narrative.

Boundary held: **swap call sites only** — no change to `batch_read_node_props` or `read_col_slot` semantics, which reach every reader in the storage layer.

## Required test

The missing one is destination-side: a one-hop directed query filtering on the *destination* with a null-bound `$param`, on a fixture with **enough neighbours to actually traverse the batch path** — plus a companion asserting a genuinely stored `0` is still found and still compares equal to `0`.

## Credit where it's due

Reusing the existing `read_node_props` helper instead of reinventing #522's inline pattern, isolating each site in its own test, exact-set assertions covering the over-match direction, and **refusing to fabricate a test for a branch that had been proved unreachable** — that last decision is the only reason this was caught at review rather than after merge. An agent that had written a passing test for the dead branch would have produced a green, plausible, wrong PR.

ryaker (commented):
## Outside review of #526 — the fix is correct, and it fixes something worse than the issue describes

Reviewed cold at `a0e279f`, with `origin/main` (`19d62c9`) as the control. Both runs from a detached worktree with its own `CARGO_TARGET_DIR`.

### The defect direction in the issue text is inverted, and the real one is more severe

#521 and this PR describe the zero sentinel as *discarding a genuinely stored `Int64(0)` as absent* — an under-read. I could not reproduce that. What `main` actually does is the opposite: it **fabricates a value for a property that was never set**.

Same test, both trees. Fixture: `A`, `B`, `C` each with a real stored `score: 0`; `D` with **no `score` at all**; edges `A→B`, `B→C`, `A→D`.

**`origin/main` (pre-fix):**

    1-hop dst score        [[B, Int64(0)], [D, Int64(0)]]        <- D has no score
    undirected score=0     [... [A, D] ...]                      <- {score:0} MATCHED D

**`a0e279f` (this PR):**

    1-hop dst score        [[B, Int64(0)], [D, Null]]            <- correct
    undirected score=0     [A,B] [B,A] [B,C] [C,B] [D,A]         <- D never appears as b

So on `main`, `RETURN b.score` returns `Int64(0)` for a node that never had the property — fabricated data, indistinguishable from a real zero — and `MATCH (a)-[]-(b {score: 0})` **matches a node lacking the property entirely**. That second one is an over-matching filter, which is the #467 class, not the under-match #521 was filed for. Note `B`'s genuinely stored `0` reads back as `Int64(0)` on *both* trees, so the "stored zero gets discarded" symptom did not occur on any path my queries reach.

This doesn't change whether to merge — the PR fixes it either way, and fixes it correctly. It changes what the issue and commit message should say, because "absent gets fabricated as 0 and over-matches a `= 0` filter" is a materially worse defect than "a stored 0 is dropped", and anyone reading the history to judge severity will get the wrong end of it.

### Item 2 — the changed test expectation is correct. Derived blind.

`undirected_backward_b_role_filter_matches_absent_node` gained a row, and a changed expectation is exactly where a fix can freeze a new bug in as intended behaviour. I derived it from the fixture and query alone, before reading the assertion or any explanatory comment:

> `$t` is null, so `{tag: $t}` admits only nodes lacking `tag` → `b` = Alice alone. `-[r]-` is undirected, so every edge incident to Alice counts in either direction: `Alice→Bob` gives `a`=Bob, `Carol→Alice` gives `a`=Carol. No self-loops. Two rows: (Bob, Alice), (Carol, Alice).

That matches the committed assertion exactly, `ORDER BY a.name` included. Independent confirmation rather than agreement — I had not read the author's derivation.

### Conversion completeness — checked, and it is complete

Every `get_node_raw` and bare `batch_read_node_props` in `hop.rs` is now a comment referencing the old behaviour; no live call site survives. Grepped for surviving zero-sentinel filters in every spelling I could think of (`filter(.. v != 0)`, `!=0`, `v > 0`, `retain(.. != 0)`) — the only hit is the explanatory comment at line 1658.

### Not verified

**Item 3 stands open.** Batch-path coverage is still asserted by construction — the `unique_dst_slots.len() == 4` argument is reasonable but it remains an argument, and I did not make the tests prove which path executed. That is the one thing I would still want before calling the coverage settled.

I also did not attempt the near-miss check that mattered most on the first revision — whether any *newly* added test exercises a branch that is itself unreachable. Given that the first version fixed a provably-dead site and the headline query stayed broken, a test asserting correct behaviour on a dead branch would look identical to a real one from the outside.

PR #527, opened 2026-08-16
fix(sparrowdb): reject $-prefixed parameter keys in execute_with_params

### LINKED ISSUE #523: execute_with_params: a mis-keyed parameter silently returns zero rows instead of erroring
Found during peer review of #522 — the reviewer's own harness hit it and briefly produced a false "your fix doesn't work" finding before they caught it. That is the tell for how easily this misleads.

**Confirmed against the published `sparrowdb@0.1.26` npm package** (Node binding, `executeWithParams`):

```
CREATE (n:Item {name:'a', tag:'x'})
query: MATCH (n:Item {tag: $t}) RETURN n.name

params {t: 'x'}     ->  [{"n.name":"a"}]     correct
params {'$t': 'x'}  ->  []                   silently empty
params {tt: 'x'}    ->  []                   silently empty
params {}           ->  []                   silently empty
```

The public API takes keys **without** the sigil (`"t"`); the engine keys them `"$t"` internally. Passing the sigil form — a natural thing to write, since that is how the parameter appears in the query text — returns an empty result set and **no error**. Same for a typo'd key, and for omitting the parameter entirely.

## Why this matters more than a usability nit

A filter API that returns zero rows for a mis-keyed parameter is **indistinguishable from a query that legitimately matched nothing.** The caller gets no signal. This is the same fail-silent shape as the cluster of bugs fixed this week (#444, #459, #467, #472, #479, #515) — the system returning a plausible empty/null answer for input it did not understand, rather than declining.

It is also actively hazardous during debugging: it cost a reviewer a false negative on a *correct* fix, and would have produced a bogus bug report if they had not run a baseline first.

## Suggested fix

Reject unknown parameter keys, or at minimum reject keys that arrive with a leading `$`, since that form is unambiguously a caller mistake and cannot be a legitimate name. An error naming the expected form (`"t"`, not `"$t"`) costs nothing and removes the whole class.

Related: the omitted-parameter case (`params {}`) is #467/#471 territory and now fails **closed** on the read path, which is correct — but the caller still gets no signal that a filter term was unresolvable. Worth considering whether an unresolvable filter should be an error rather than an empty result, at least for parameters that were never supplied.

Reachable from every binding (`executeWithParams` in Node, equivalents in Python/Ruby/Rust), not specific to one.

ryaker:
## Summary
Closes #523.

`GraphDb::execute_with_params` takes a `HashMap<String, Value>` keyed by the *bare* parameter name (the Cypher lexer strips the `$` sigil when tokenizing `$name` in query text, so `Expr`/`Literal::Param` and the engine's lookup never see a leading `$`). A caller who mirrors the query text and keys the map `"$t"` instead of `"t"` was previously silently mis-keyed: the lookup missed, the filter fell back to unresolvable, and (post-#467) that fails **closed** — zero rows, no error. That's indistinguishable from a query that legitimately matched nothing, and it cost a reviewer a false "fix doesn't work" finding on correct code.

**Root cause:** `execute_with_params` (`crates/sparrowdb/src/db.rs`) had no validation on the incoming params-map keys before handing them to the engine's bare-name lookup.

**Fix:** `execute_with_params` now rejects any params-map key with a leading `$` up front, with an error naming both the offending key and the expected bare form. This is the single entry point shared by every binding (Node's `executeWithParams`, Python's `execute_with_params`, the HTTP `/cypher` handler, and `helpers.rs`), so the fix is inherited by all of them with **zero binding changes**.

**Function changed:** `GraphDb::execute_with_params` in `crates/sparrowdb/src/db.rs` only — a validation loop added at the top of the function, before parsing/binding. No other function touched.

## Shared-param-map pattern — investigated, not found in this repo
The issue asked me to check whether callers pass a shared param map across several queries (which unknown-key rejection could break) before deciding whether to also reject unreferenced/unknown keys. I checked every `execute_with_params` call site in this repo (Node/Python bindings, HTTP server, `helpers.rs`, and all test fixtures) — every one builds a params map tailored to exactly the query it's passed to; none reuse a map with extra keys across calls. That's not proof no external caller does this, but I found no in-repo evidence either way.

Given that, and given that full unknown-key detection requires walking the entire bound AST for every `$param` reference (`CASE` branches, list comprehensions, subqueries, function args, ...) — the same kind of AST-completeness problem that caused #478 and #502 when a `Statement`/`Expr` variant fell through a match that wasn't kept in sync — I implemented **only** the `$`-prefix rejection here. It's unambiguous (no legitimate parameter name can contain `$`) and complete by construction. A typo'd key (`tt` for `t`) still returns silent-empty; closing that gap is a separate, larger change I'm leaving for a follow-up decision rather than bundling into this fix.

## Deliberately out of scope
The omitted-parameter case (`params {}`) is unchanged — still fails closed (empty rows, no error), which is correct #467/#471 behavior. `regression_467.rs`'s existing tests for this pass unmodified. Whether an *unresolvable* filter (as opposed to a caller mistake) should itself be an error is a separate design decision, called out in the issue as the maintainer's call, not mine.

## Test plan
- [x] New file `crates/sparrowdb/tests/regression_523_param_key_sigil.rs`, 5 tests: correct bare key still matches; sigil-prefixed key now errors (naming both `"$t"` and `"t"` in the message); typo'd key still silent-empty (control, explicitly out of scope); omitted param still silent-empty (control, #467/#471); sigil key errors even when a sibling key in the same map is correctly formed.
- [x] Confirmed each new test **fails against the pre-fix commit** (stashed the `db.rs` change, kept the new test file, ran it): both the sigil-key tests failed with the exact reported symptom — `rows: []`, `Ok(...)`, no error — while the three controls passed. Restored the fix afterward.
- [x] Verified through the **Node binding** (`executeWithParams`), not just the Rust API — built `sparrowdb-node` release, copied the `.node` artifact into `npm/sparrowdb/`, and reran the issue's exact repro:
  - `{t: 'x'}` → `[{"n.name":"a"}]` (unchanged, correct)
  - `{'$t': 'x'}` → now **throws**: `invalid argument: parameter key "$t" must not include the leading '$' sigil; use "t" instead — ...`
  - `{tt: 'x'}` → `[]` (unchanged, out of scope)
  - `{}` → `[]` (unchanged, control)
  Since the fix is in `GraphDb::execute_with_params` (shared code), Python's `execute_with_params`, the HTTP server's `/cypher` handler, and `helpers.rs` all inherit it automatically — I did not build/test those bindings individually, but they route through the identical entry point verified above.
- [x] `cargo fmt` — clean, no diff.
- [x] `cargo clippy -p sparrowdb -p sparrowdb-execution --all-targets --keep-going` — no warnings.
- [x] `regression_467` (11 tests) and `regression_472_479` (9 tests) — both pass unmodified, confirming this change doesn't disturb `$param` filter fail-closed behavior.
- [ ] Full `cargo test -p sparrowdb --no-fail-fast` — running in the background at commit time (suite is slow, 60+ min); not blocking this PR per repo convention, CI is authoritative.

## Could not verify
- Python and Ruby bindings were **not** individually built/tested — I relied on the fact that they call the same `GraphDb::execute_with_params` verified directly and via Node, and confirmed by reading each binding's source that none re-implement the params-map conversion in a way that would bypass this check. Ruby doesn't have an `execute_with_params` binding at all currently, so it's unaffected either way.
- Whether an external caller (outside this repo) actually relies on a shared/reused param map with unreferenced keys — I found no such pattern in this codebase, but that's not proof none exists elsewhere. This is why I scoped the fix to the unambiguous `$`-prefix case only, per the issue's "at minimum" instruction.
- The full workspace suite (`cargo test -p sparrowdb --no-fail-fast`) was still running in the background when this PR was opened; will report results in a follow-up comment if anything unexpected turns up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Parameter keys beginning with `$` now return a clear validation error instructing you to use the bare parameter name.
  * Valid parameter keys continue to work, while unused or misspelled keys retain their existing behavior.
  * Missing parameters continue to fail safely.

### DIFF
--- a/crates/sparrowdb/src/db.rs
+++ b/crates/sparrowdb/src/db.rs
@@ -1233,7 +1233,22 @@ impl GraphDb {
         // branches, list comprehensions, subqueries, ...) and is left as a
         // separate, larger change.
         for key in params.keys() {
-            if let Some(bare) = key.strip_prefix('$') {
+            if key.starts_with('$') {
+                // `trim_start_matches`, not `strip_prefix`: a key can carry
+                // more than one leading sigil (`"$$t"`), and suggesting the
+                // single-strip result (`"$t"`) would hand back a value this
+                // same check rejects on the very next call — an error whose
+                // own advice fails.
+                let bare = key.trim_start_matches('$');
+                if bare.is_empty() {
+                    // The key was only sigils (`"$"`, `"$$"`) — there is no
+                    // bare name left to suggest.
+                    return Err(Error::InvalidArgument(format!(
+                        "parameter key {key:?} has no name left once its \
+                         leading '$' is removed; the params map must be \
+                         keyed by a real parameter name, not the sigil alone"
+                    )));
+                }
                 return Err(Error::InvalidArgument(format!(
                     "parameter key {key:?} must not include the leading '$' \
                      sigil; use {bare:?} instead — the params map is keyed by \
--- a/crates/sparrowdb/src/db.rs
+++ b/crates/sparrowdb/src/db.rs
@@ -1210,6 +1210,39 @@ impl GraphDb {
         cypher: &str,
         params: HashMap<String, sparrowdb_execution::Value>,
     ) -> Result<QueryResult> {
+        // SPA-523: the params map is keyed by the bare parameter name — the
+        // lexer strips the `$` sigil when tokenizing `$name` inside Cypher
+        // source (see `sparrowdb_cypher::lexer`), so `Expr`/`Literal::Param`
+        // and the engine's lookup (`params.get(name)`, `EngineBuilder::
+        // with_params`) never see a leading `$`. A caller who mirrors the
+        // query text and keys the map `"$t"` instead of `"t"` was previously
+        // silently mis-keyed: the lookup missed, the filter fell back to an
+        // unresolvable value, and (post-#467) that now fails *closed* —
+        // zero rows, no error. That is indistinguishable from a query that
+        // legitimately matched nothing. Reject the sigil form outright: it
+        // is unambiguously a caller mistake (no legitimate parameter name
+        // can contain `$`, since the lexer would never let one reach the
+        // caller's params map to begin with) and the error can name the
+        // exact fix.
+        //
+        // Unknown-but-unreferenced keys (a key present in the map that no
+        // `$name` in the query text refers to) are deliberately NOT rejected
+        // here — see issue #523 discussion. A typo'd key (`tt` for `t`) still
+        // returns silent-empty rather than erroring; catching that requires
+        // walking the full bound AST for every `$param` reference (CASE
+        // branches, list comprehensions, subqueries, ...) and is left as a
+        // separate, larger change.
+        for key in params.keys() {
+            if let Some(bare) = key.strip_prefix('$') {
+                return Err(Error::InvalidArgument(format!(
+                    "parameter key {key:?} must not include the leading '$' \
+                     sigil; use {bare:?} instead — the params map is keyed by \
+                     the bare name (e.g. `{{{bare:?}: ...}}`), and `$` is only \
+                     written in the Cypher query text itself (e.g. `${bare}`)"
+                )));
+            }
+        }
+
         use sparrowdb_cypher::{bind, parse};
 
         let stmt = parse(cypher)?;
--- a/crates/sparrowdb/tests/regression_523_param_key_sigil.rs
+++ b/crates/sparrowdb/tests/regression_523_param_key_sigil.rs
@@ -1,0 +1,173 @@
+//! Regression guard for issue #523 — a mis-keyed parameter silently returns
+//! zero rows instead of erroring.
+//!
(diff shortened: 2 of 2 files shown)


ryaker:
Merging. CodeRabbit's finding is addressed — `trim_start_matches('$')` now strips all leading sigils so `"$$t"` recommends `"t"`, and a key that is only sigils gets its own distinct error rather than suggesting an empty name. Verified on the pushed branch.

Review state: the **first** pass was genuine (claude-review 15 turns/$3.14, CodeRabbit a real review — the first PR in two days where both actually ran), and that is where the finding came from. The re-review after the fix push no-opped (4 turns/$0.156), so treat the second run as unverified; the substantive review happened on the earlier commit.

Scoping to $-prefix rejection only was the right call and for the right reason: rejecting *unknown* keys requires walking the entire bound AST for every `$param` reference — CASE branches, list comprehensions, subqueries, function args — which is the same completeness trap that produced #478 and #502 here. Tracked as a follow-up decision rather than half-done.

Issue #462, opened 2026-08-02
CRITICAL: FTS index errors are swallowed on both the write and read paths — documents silently unsearchable, queries silently answer 'no match'


ryaker:
The same defect class as #445/#446 on the vector side, still open on the full-text side. Every line below verified against `origin/main` at `5f36b02`.

## 1. WRITE PATH — the serious one. `crates/sparrowdb/src/db.rs:984-986`

```rust
if let Ok(text) = std::str::from_utf8(bytes) {
    if let Ok(mut idx) = FtsIndex::open(&self.inner.path, &label, prop_name) {
        idx.insert(node_id.0, text);
        if let Err(e) = idx.save() {
            warn!("FTS index save failed for ({label}, {prop_name}): {e}");
        }
    }
}
```

If `FtsIndex::open` fails, **the text is silently never indexed**. The node is written and committed; the document simply never becomes searchable, forever, with no error and no warning. Only a later full reindex would recover it, and nothing tells anyone to run one.

**The inconsistency is the tell**: a failing `save()` is warned about, a failing `open()` is not. The code already accepts that FTS failures are worth surfacing — this path just misses the earlier one. Note too that a non-UTF-8 `from_utf8` failure is swallowed by the same `if let`.

This is exactly #445: *"a damaged index does not surface as an error — it surfaces as 'there is no index here', and every write is silently dropped."*

## 2. READ PATH — `crates/sparrowdb-execution/src/engine/mod.rs:274-282`

```rust
let mut cache = self.fts_cache.lock().ok()?;       // poisoned mutex -> None
...
match FtsIndex::open(&self.db_root, label, property) {
    Ok(idx)  => { cache.insert(key.clone(), idx); }
    Err(_)   => return None,                        // damaged index -> None
}
```

`None` collapses three distinct states into one: **no index configured**, **index damaged**, and **another thread panicked while holding the lock**.

Its callers then turn `None` into a confident wrong answer:

| caller | `None` becomes | meaning to the user |
|---|---|---|
| `expr.rs:151` (`matches_query`) | `Value::Bool(false)` | "this document does not match" |
| `expr.rs:208` (`bm25_score`) | `Value::Float64(0.0)` | "relevance is zero" |

A damaged FTS index therefore reports every document as a non-match with a score of 0.0. Indistinguishable from a genuinely empty result.

## 3. `hybrid_search`'s own FTS branch — `crates/sparrowdb-execution/src/engine/expr.rs:472-478`

```rust
[...]

PR #528, opened 2026-08-16
fix(npm): reconcile index.d.ts with its napi generator; de-export PLATFORM_BINARIES

### LINKED ISSUE #513: index.d.ts has drifted from its NAPI generator — a regen reintroduces already-fixed bugs

### LINKED ISSUE #518: npm: PLATFORM_BINARIES is exported as public API purely for test access
Minor cleanup, deliberately deferred out of #517 so an urgent fix could ship.

#517 needed the platform→filename map readable from `test/platform-resolution.test.js`, because a test that re-derives the naming rule drifts from it — that was the actual defect its first CI run caught (`sparrowdb.linux-x64.node` vs the real `sparrowdb.linux-x64-gnu.node`). Reading the map instead of reconstructing it was the right call.

The implementation exposes it on the package's public surface:

```js
// npm/sparrowdb/index.js
module.exports.PLATFORM_BINARIES = PLATFORM_BINARIES
```

So `require('sparrowdb').PLATFORM_BINARIES` now works for any consumer. It is commented as not-public-API, but a comment does not stop anyone depending on it, and removing it later would technically be a breaking change.

## Suggested fix

Move the map to its own tiny module — `npm/sparrowdb/platforms.js` — and have both `index.js` and the test `require` it. Same single-source property, zero public surface.

## Why this was not blocking

- The behaviour is correct and verified end-to-end on darwin-arm64 (both the load path and the wrong-binary error path).
- #517 fixes a break where **`npm install sparrowdb` has never worked on macOS ARM** — 0.1.20 through 0.1.25 all fail with `ERR_DLOPEN_FAILED`. Getting that out mattered more than a third review round-trip on an export nit.
- The leak is low-risk in practice: it is a frozen-in-spirit map of two entries, and it does not appear in `index.d.ts`, so TypeScript consumers cannot discover it.

Worth doing before it accretes a dependent, not worth having delayed the fix.

ryaker:
## Summary

Fixes two npm-packaging issues, both scoped to `npm/sparrowdb/` and `crates/sparrowdb-node/src/lib.rs`.

closes #513, closes #518

### #513 — index.d.ts had drifted from its napi-rs generator

Ran the actual regen procedure (`napi build --release --platform --dts <scratch> --js false --cargo-cwd ../../crates/sparrowdb-node --cargo-name sparrowdb_node`) to a scratch path and diffed against the checked-in `index.d.ts`. Every difference fell into one of three buckets:

1. **A real napi-derive-backend bug**, not a hand-fix gone stale: `static open(path: string): SparrowDb` (wrong casing). `napi-derive-backend`'s `gen_ts_func_ret()` derives a `#[napi(factory)]` method's TS return type from the raw Rust struct ident via `to_case(Case::Pascal)` — it never consults `#[napi(js_name = "SparrowDB")]`, so "DB" gets re-cased to "Db" as an all-caps-acronym word. The #405 hand-fix patched the *artifact* but a regen would silently reintroduce the bug. Fixed at the source with `#[napi(factory, ts_return_type = "SparrowDB")]`, which overrides the derived return type outright (confirmed against `napi-derive-2.16.13`'s attribute parser — `ts_return_type` is a real, valid fn-level override).
2. **Artifact ahead of source**: the "HNSW note (as of 0.1.23)" paragraph on `executeWithParams` existed only in the checked-in `.d.ts`. Traced via `git log -S` — it was hand-added to `index.d.ts` in #410 despite that commit's message claiming a regen; it was never in `lib.rs`. Moved into the Rust doc comment so it survives a regen. Also synced "struct doc comment" → "interface doc comment" (correct for TS-facing prose) from the artifact into `lib.rs`.
3. **Source ahead of artifact**: doc-comment prose for #441 (`addToVectorIndex` now replaces rather than no-ops, and writes the node property as a recovery path) and #462 (`hybridSearch` distinguishes a damaged FTS index from a merely-unconfigured one) existed in `lib.rs` but had never been synced into the checked-in `.d.ts`. Took the regenerated text.

Result: `tsc -p tsconfig.json` passes. `kind` (hand-patched into `VectorIndexLoadFailure` in #520) is confirmed present and now reproduced by its own `lib.rs` doc comment — no longer at risk from a future regen.

### #518 — PLATFORM_BINARIES exported on the public API purely for test access

`npm/sparrowdb/index.js` exported the platform→filename map so `test/platform-resolution.test.js` could read it instead of re-deriving the naming convention (the re-derivation bug is what broke #517's first CI run: `sparrowdb.linux-x64.node` vs the real `sparrowdb.linux-x64-gnu.node`). Moved the map to `npm/sparrowdb/platforms.js`, required by both `index.js` and the test — same single source, zero public surface. Added to `package.json`'s `files` array so it ships in the tarball, and `test/platform-resolution.test.js`'s `makeIsolatedPackageDir()` now copies it alongside `index.js`.

## Verification

- **Regen diff**: full diff shown above, every hunk accounted for.
- **`tsc -p tsconfig.json`**: passes clean.
- **`kind` field survival**: confirmed present in regenerated output at the correct location, `grep -c "SparrowDb\b"` (bare, wrong casing) = 0, `SparrowDB\b` = 18.
- **`npm test`** (`node --test test/*.test.js`): 50/50 pass, 0 failures — using a binary built fresh in this task's own isolated `CARGO_TARGET_DIR=/Volumes/Dev/target-npm` (not the shared/contaminated cache). An earlier run against a borrowed binary from the shared cache showed 3 unrelated failures in `integration.test.js` (vector-index/param tests); those are not present with a clean build and were never touched by this PR.
- **`npm pack --dry-run`**: 4 files — `index.js`, `index.d.ts`, `package.json`, `platforms.js` (no stray `.node`, confirmed after removing a leftover local build artifact from my own regen runs).
- **`platforms.js` not picked up as a test**: `npm test` globs `test/*.test.js`; `platforms.js` lives at package root.
- **Unsupported-platform path forced manually** (`process.platform='win32'`, `arch='x64'`): throws a catchable `Error` (not a bare `MODULE_NOT_FOUND`) with the clear "no prebuilt native module" message.
- **Regression-detector still works**: proved the naive-reconstruction bug pattern from #481/#517 (`` `sparrowdb.${platform}-${arch}.node` ``) mismatches on `linux-x64` (`sparrowdb.linux-x64.node` vs real `sparrowdb.linux-x64-gnu.node`) while coincidentally matching on `darwin-arm64` — exactly the original failure mode. Since both `index.js` and the test now `require()` the identical `platforms.js`, this class of drift is structurally impossible, not just avoided by convention.
- **`cargo check -p sparrowdb-node`**, **`cargo clippy -p sparrowdb-node --no-deps`**, **`cargo fmt --check`**: all clean.
- **Prettier**: repo has no `.prettierrc`, so `prettier --write` would reformat to double-quotes/semicolons against the existing single-quote/no-semi style (same issue #517 hit). Left unformatted; only the functional diff was hand-applied.

### Could not verify

- Did not test on an actual `linux-x64` machine/CI — verification of the `-gnu` suffix correctness relies on reading `platforms.js`'s value and CI's own pass/fail, not a local linux run (this machine is darwin-arm64).
- Did not run the full `cargo test --workspace` suite (excluded per repo `CLAUDE.md`: `sparrowdb-ruby` doesn't compile locally) — only `cargo check`/`clippy`/`fmt` on the touched crate, since the scope here is npm packaging, not engine behavior.
- Have not run this against a real `npm publish --dry-run` with 2FA/registry auth — no publish was attempted per instructions.
- Did not add the CI "regenerate-and-assert-no-diff" check that issue #513 calls "the real deliverable" for preventing *future* drift — out of scope for what was asked (regen-and-reconcile + verification), flagging as a natural follow-up.

**Not merged** — left for review per instructions.

### DIFF
--- a/crates/sparrowdb-node/src/lib.rs
+++ b/crates/sparrowdb-node/src/lib.rs
@@ -334,8 +334,8 @@ pub struct VectorIndexLoadFailure {
     /// not authoritative and not machine-parseable.
     pub reason: String,
     /// `"live_unserviceable"` or `"quarantined_no_live_index"` — see the
-    /// struct doc comment. Check this before deciding whether a vector write
-    /// against this `(label, prop)` would currently succeed or throw.
+    /// interface doc comment. Check this before deciding whether a vector
+    /// write against this `(label, prop)` would currently succeed or throw.
     pub kind: String,
 }
 
@@ -469,7 +469,16 @@ impl SparrowDB {
     /// Open (or create) a SparrowDB database at `path`.
     ///
     /// Throws if the path cannot be created or the database files are corrupt.
-    #[napi(factory)]
+    //
+    // `ts_return_type` below is required, not decorative: napi-derive-backend
+    // derives a factory method's TS return type from the raw Rust struct
+    // ident via `to_case(Case::Pascal)`, which does not consult this struct's
+    // `#[napi(js_name = "SparrowDB")]` and instead treats "DB" as an all-caps
+    // acronym word, re-casing it to "Db". Without the override below, a
+    // regenerated `index.d.ts` reads `static open(...): SparrowDb` — the
+    // exact factory-casing bug fixed by hand in #405. `js_name` alone only
+    // controls the `export declare class SparrowDB` line.
+    #[napi(factory, ts_return_type = "SparrowDB")]
     pub fn open(path: String) -> napi::Result<Self> {
         let db = ::sparrowdb::GraphDb::open(std::path::Path::new(&path)).map_err(to_napi)?;
         Ok(SparrowDB { inner: db })
@@ -622,6 +631,12 @@ impl SparrowDB {
     /// embeddings must be passed as parameters.  The engine surface
     /// (`GraphDb::execute_with_params`) has supported this since SPA-218; this
     /// binding simply exposes it to JS callers.
+    ///
+    /// **HNSW note (as of 0.1.23):** when the target property has a vector
+    /// index, `SET n.prop = $vec` now also writes the vector into the HNSW
+    /// index. Previously the property was stored but the HNSW file was not
+    /// updated — a silent data-loss bug surfaced by KMSmcp channel message
+    /// #202 (#410).
     ///
     /// **`CREATE` support (SPA-480):** standalone `CREATE` and
     /// `MATCH ... CREATE` — including edge properties — now accept `$param`
--- a/npm/sparrowdb/index.d.ts
+++ b/npm/sparrowdb/index.d.ts
@@ -122,8 +122,8 @@ export interface VectorIndexLoadFailure {
   reason: string
   /**
    * `"live_unserviceable"` or `"quarantined_no_live_index"` — see the
-   * interface doc comment. Check this before deciding whether a vector write
-   * against this `(label, prop)` would currently succeed or throw.
+   * interface doc comment. Check this before deciding whether a vector
+   * write against this `(label, prop)` would currently succeed or throw.
    */
   kind: string
 }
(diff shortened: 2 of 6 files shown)


ryaker:
Merging, with the review state on the record: **CodeRabbit was rate-limited and claude-review was thin** (8 turns / $0.22 — a genuine deep pass on a sibling PR was 15 turns / $3.14). So CI plus my own verification is what this rests on.

**What I verified independently — the durable claim in #513:**

The casing fix is now **at the source**, not patched into the artifact for a third time:

```rust
#[napi(factory, ts_return_type = "SparrowDB")]
pub fn open(path: String) -> napi::Result<Self>
```

with a comment explaining why it is load-bearing rather than decorative. Confirmed in the committed `.d.ts`: `SparrowDb` (bad casing) appears **0** times, `SparrowDB` 18 times, and line 218 reads `static open(path: string): SparrowDB`. A regen will now reproduce that.

That distinction is the whole point of #513. #405 fixed the symptom in the generated file, so every regen silently reintroduced it. Tracing it into `napi-derive-backend`'s `gen_ts_func_ret()` — which derives a factory's return type from the raw struct ident via `to_case(Case::Pascal)` and never consults `js_name` — is what turned a recurring hand-patch into a fix.

**#518** is resolved structurally: both `index.js` and the test now `require` the same `platforms.js`, so they cannot disagree about a filename. That closes the failure mode that broke #517's first CI run, where a test re-derived `sparrowdb.linux-x64.node` against the real `sparrowdb.linux-x64-**gnu**.node` and passed on darwin only by coincidence.

Also worth recording: an earlier test run borrowed a binary from the shared `/Volumes/Dev/cargo-target` cache and showed 3 unrelated failures. That result was correctly discarded as contaminated and the binary rebuilt in an isolated `CARGO_TARGET_DIR` — which is exactly the rule in this repo's CLAUDE.md, applied without being reminded.

**Not done, flagged as follow-up:** the CI "regenerate and assert no diff" check that would prevent future drift mechanically. Worth filing.

Issue #452, opened 2026-08-02
CRITICAL: #442 introduced a new data-loss surface — deterministic .tmp path collides between writers, loser gets ENOENT that is_lost_update() cannot recognise


ryaker:
## This is a regression introduced by #442, not a pre-existing bug

Before #442 there was no temp file at all — `save()` was a bare `fs::write` to the real path. #442 added crash-safety via temp + fsync + rename, and `temp_path()` (`vector_index.rs:669-673`) is a **deterministic `<path>.tmp`** with no pid, no nonce, no uniqueness:

```rust
fn temp_path(path: &Path) -> PathBuf {
    let mut s = path.to_path_buf().into_os_string();
    s.push(".tmp");
    PathBuf::from(s)
}
```

Two concurrent writers both `File::create` the same temp path. The first to `rename` vacates it; the loser's `rename` fails with **`ENOENT` — "No such file or directory (os error 2)"**, and its data is gone.

**Reproduced 27 of 27** barrier-synchronised runs by the affected consumer: exactly one `Ok`, one `ENOENT`, **never** the advertised generation-conflict error.

## Why it is worse than plain data loss

`is_lost_update()` (`vector_index.rs:381`) string-matches the generation-conflict prefix:

```rust
pub fn is_lost_update(err: &std::io::Error) -> bool {
    err.to_string().starts_with(LOST_UPDATE_PREFIX)
}
```

The ENOENT variant cannot satisfy it. So a consumer following the documented recovery contract — catch lost-update, reopen, retry — reads a **data-loss event as a disk fault** and does not retry. #442's own error-handling contract is unreachable on the path that actually fires.

## Fix

`<path>.tmp.<pid>.<nonce>`. Small, and it closes a surface that #442 itself opened. This is a strict regression against pre-#442 behaviour and should land before anything else in this series.

## The remaining TOCTOU is separate and not closed by the above

`save()` (`vector_index.rs:309-377`) reads the on-disk generation, then serialises, then writes + renames, with **no lock across that window**. Two processes that both load generation N both pass the check. Unique temp paths remove the ENOENT collision but leave this: both saves return `Ok`, and the second silently clobbers the first — the original #441 failure mode, **narrowed from unbounded to milliseconds, not eliminated**.

Every `insert()` calls `save()`, so a multi-thousand-vector backfill opens that window thousands of times. This needs a real exclusion mechanism (advisory lock, or O_EXCL on a lock file), not a tighter check.

## The existing test cannot catch either

[...]

PR #531, opened 2026-08-17
fix(db): take exclusive cross-process lock at open() (closes #524)

### LINKED ISSUE #524: CRITICAL: two processes writing concurrently permanently corrupt the catalog — database becomes unopenable
**Two processes creating labels concurrently on the same database root permanently corrupt `catalog.tlv`. The database then cannot be opened at all.**

```
Error: corruption: duplicate label_id 0 in catalog file
```

Reproduced against the published `sparrowdb@0.1.26` npm package on darwin-arm64, Node 22.22.0. **Not a branch artifact — this is shipped.**

## Reproduction

`labelrace.js`:
```js
const { SparrowDB } = require('sparrowdb')
const db = SparrowDB.open(process.argv[2])
db.execute(`CREATE (n:${process.argv[3]} {id:'${process.argv[3]}'})`)
```

```bash
D=$(mktemp -d)
node labelrace.js "$D" Alpha & node labelrace.js "$D" Beta & wait
node -e "require('sparrowdb').SparrowDB.open('$D')"
```

**Result: 4 of 5 runs brick the database.**

```
run 1: BRICKED — corruption: duplicate label_id 0 in catalog file
run 2: BRICKED — corruption: duplicate label_id 0 in catalog file
run 3: BRICKED — corruption: duplicate label_id 0 in catalog file
run 4: BRICKED — corruption: duplicate label_id 0 in catalog file
run 5: OPENS OK — [{"n.id":"Beta"},{"n.id":"Alpha"}]
```

The 5th run is the interleaving that happened not to collide.

## Bounded

| access pattern | result |
|---|---|
| single process | fine |
| **sequential** multi-process (one exits, then the next opens) | **fine, 3/3** |
| **concurrent** multi-process | **corrupt, 4/5** |

So it is purely a race, not a general multi-process incompatibility.

## Cause

`db.rs:792` already states it: *"Both paths derive `next_label_id` from their own in-memory counter, so concurrent catalog writes without this guard can assign duplicate IDs."* The guard is `WriteGuard::try_acquire` — **in-process only**.

`GraphDb::open()` places **no cross-process lock**. A freshly created database directory contains only:

```
catalog.tlv   edges   nodes   wal
```

No lock file, no advisory lock on open. So two processes each open successfully, each allocate `label_id` from an independent counter starting at 0, and both write id 0 into `catalog.tlv`.

## Severity

This destroys a database rather than returning a wrong answer, and recovery requires hand-editing or discarding `catalog.tlv`. Any deployment where two processes touch one database root is exposed — a daemon plus a CLI invocation, two workers, a cron job overlapping a service, or a health check that opens the store. Note the binding exposes only `SparrowDB.open()` with **no read-only mode**, so *every* attach is a potential writer.

There is a corruption *detector* — it fails closed on read, which is right — but nothing **prevents** the write.

## Suggested direction

A cross-process advisory lock (`flock` on a lockfile in the db root) taken at `open()`, with a clear error when another process holds it. If concurrent multi-process access is intended to be unsupported, that should be enforced at `open()` and documented, rather than permitted and left to corrupt silently.

## Provenance

Found via cross-session peer review of #520. The reviewer raised duplicate `label_id` as an explicit **hypothesis with a named experiment** — *"two processes, two `create_label` calls interleaved, then check whether the catalog holds two names at one id"* — and was careful to say it had not reproduced it and should not be treated as a finding. I ran the experiment. It reproduces, and the consequence is worse than hypothesised: not a mis-routed write to a wrong-but-real `(label, prop)` key, but an unopenable database.

ryaker:
## Root cause

Two processes opening the same database root each derive catalog counters
(`next_label_id`, etc.) from independent in-memory state, with nothing
coordinating them across process boundaries. Concurrent `CREATE`s of a new
label can both allocate `label_id` 0, corrupting `catalog.tlv` beyond what
`open()` will parse again:

```
Error: corruption: duplicate label_id 0 in catalog file
```

`GraphDb::open` took no lock at all before this change — a freshly created
db root had no lock file, nothing preventing a second process from
attaching mid-write.

**Reproduced against pre-fix `origin/main` (4d15340)**, built via
`sparrowdb-cli` and run as two racing OS processes (mirrors the Node.js
repro from the issue):

```
run 1: BRICKED -> Error: corruption: duplicate label_id 0 in catalog file
run 2: BRICKED -> Error: corruption: duplicate label_id 0 in catalog file
run 3: BRICKED -> Error: corruption: duplicate label_id 0 in catalog file
run 4: BRICKED -> Error: corruption: duplicate label_id 0 in catalog file
run 5: BRICKED -> Error: corruption: duplicate label_id 0 in catalog file
run 6: BRICKED -> Error: corruption: duplicate label_id 0 in catalog file
run 7: BRICKED -> Error: corruption: duplicate label_id 0 in catalog file
run 8: BRICKED -> Error: corruption: duplicate label_id 0 in catalog file
```

8/8 here (vs. the issue's 4/5) — the CLI pays process-startup cost before
`open()`, which widens the race window relative to the in-process Node repro.

## The fix

`GraphDb::open`/`open_encrypted` now take an exclusive `flock` on
`<db_root>/db.lock` (`ProcessLock`, new module `crates/sparrowdb/src/process_lock.rs`)
before touching anything else — catalog, WAL, vector indexes. A second
process racing to open the same root gets an immediate `Error::DatabaseLocked`
instead of a chance to corrupt the catalog.

**Decision, recorded per the issue discussion**: exclusive rather than a
narrower lock scoped to catalog mutation. Every language binding (Node,
Python, Ruby) exposes only `open()` with no read-only mode today, so
"preserve concurrent readers" is not a real case being traded away — it's
aspirational regardless of what this lock does. Exclusive-at-`open()` now,
a real read-only mode later, is the honest ordering.

**Why `flock` and not a PID file**: a PID file has to answer "what if the
holder died?", and every answer is a heuristic — liveness probes race PID
reuse, age thresholds either wedge the database after a crash or break the
lock out from under a slow-but-live writer. `flock` is released by the
kernel when the file descriptor closes, including on `SIGKILL`, so a
crashed holder leaves nothing to reclaim and nothing stale to reason about.
This mirrors `SaveLock` in `sparrowdb-storage::vector_index` (#452), which
made the identical call for per-index locking.

**The one subtlety**: `flock` is scoped to the *open file description*, not
the process. A naive per-`open()` `try_lock()` would therefore also block a
second `GraphDb::open` from the *same* process — which several existing
tests intentionally rely on (e.g. `vector_index_durability`'s
`a_stale_handle_cannot_silently_revert_another_writers_vectors`, which holds
two live handles on one root to test how a stale in-process handle behaves).
#524 is about two *processes* racing an uncoordinated in-memory counter, not
about forbidding two handles in one address space. `ProcessLock` resolves
this with a process-local registry (keyed by canonicalized root) that hands
out the same `Arc` to every same-process caller — only the first caller per
root actually contends for the OS-level lock. A second, genuinely separate
process still calls `try_lock` fresh and is refused by the kernel exactly as
before.

## What changed, per file

- `crates/sparrowdb-common/src/lib.rs` — new `Error::DatabaseLocked(String)`
  variant with an actionable message (names the path, says what's holding it,
  says what to do).
- `crates/sparrowdb/src/process_lock.rs` (new) — `ProcessLock`: the flock +
  process-local sharing registry described above. 3 unit tests.
- `crates/sparrowdb/src/types.rs` — `DbInner` gains a `_db_lock: ProcessLock`
  field, held only for its `Drop` side effect.
- `crates/sparrowdb/src/db.rs` — `open()`/`open_encrypted()` acquire the lock
  as their first action, before `create_dir_all`'s directory is touched by
  anything else; updated `# Errors` doc sections; added a cross-reference
  comment on the existing in-process `WriteGuard` note (db.rs:~799) pointing
  at the new cross-process guard.
- `crates/sparrowdb/tests/regression_524_process_lock.rs` (new) — see below.
- `README.md` — the `#525` safety banner said "no lock file and no guard at
  `open()`"; that's now false, so it's rewritten to describe the lock, the
  crash-release guarantee, and that sequential access remains safe (kept the
  pre-fix corruption history as context for anyone already hit by it).

## Testing

All three properties below are exercised across **real OS processes**
(`std::process::Command` + `current_exe()` re-invocation, the same pattern
`regression_464_quarantine_race.rs` already uses in this repo), not threads
— threads would share one `DbInner`'s in-memory state and couldn't
reproduce the bug at all.

- **Concurrent case fails cleanly** — `concurrent_open_one_wins_one_gets_a_clean_error`:
  two processes race `open()` + `CREATE` against a fresh root. Exactly one
  wins (asserted via XOR, not "at least one"); the loser's error is asserted
  to both originate at `open()` (not the write) and contain "database
  locked". A subsequent open sees exactly the winner's row — no corruption,
  no lost write. Verified this test **fails against pre-fix `origin/main`**
  (copied into a worktree at 4d15340): 4/4 runs failed, symptom was the
  loser's write hitting a raw I/O race (`No such file or directory`) rather
  than the classic duplicate-`label_id` corruption — a different symptom of
  the same missing-coordination root cause, not a new bug.
- **Abnormal exit releases the lock** — `lock_is_released_when_holder_is_sigkilled`:
  holder process opens and signals ready; parent confirms a second `open()`
  is refused while it's alive; `SIGKILL`s it (`Child::kill` is SIGKILL on
  Unix); confirms the very next `open()` succeeds immediately and the db is
  fully usable (not just openable) — proving release comes from the kernel
  closing the fd, not from `ProcessLock`'s `Drop` impl (which SIGKILL skips
  entirely). Verified this **fails against pre-fix `origin/main`** at the
  "must be refused while alive" assertion (open just succeeds, no lock at
  all pre-fix).
- **Sequential access is unaffected** — `sequential_open_write_close_reopen_still_works`,
  plus the pre-existing suite: every "session 1 / session 2" reopen test in
  the crate (`spa191_rel_type_persistence`, `regression_485_stale_count`,
  `regression_491_stale_db_counts`, `spa_415_unwind_match_mutate`,
  `spa_209_schema_introspection`, `regression_436_dangling_edges`,
  `spa_184_wal_durability`, `spa_222_csr_lazy_load`,
  `regression_502_call_subquery_mutation_dispatch`) still passes.
- **Same-process multi-handle case unaffected** —
  `vector_index_durability::a_stale_handle_cannot_silently_revert_another_writers_vectors`
  still passes (this is what the process-local sharing registry exists for;
  a naive implementation breaks it).
- **No per-query cost** — the lock is acquired exactly once, inside
  `open()`/`open_encrypted()`; nothing was added to any query or write path.
- **Full suite**: `cargo test -p sparrowdb --no-fail-fast` — all binaries
  green after one fix (see "could not verify" — one flaky unit test was
  caught and fixed by this exact run). `cargo fmt --check` and
  `cargo clippy -p sparrowdb -p sparrowdb-execution --all-targets --keep-going`
  clean. `cargo check --workspace --exclude sparrowdb-ruby --all-targets`
  clean (Ruby doesn't compile locally per this repo's `CLAUDE.md`, unrelated
  to this change).

Every expected value in the new test file is derived by hand from what the
fix's contract requires (exactly one winner, exactly one row after), not
captured from a run.

## Could not verify

- **CI's Ruby job** — cannot compile locally (`RUBY_T_MOVED` / `dcompact`
  against the installed Ruby, per this repo's own `CLAUDE.md`). The `Error`
  enum change is additive and every binding (`sparrowdb-node`,
  `sparrowdb-python`, `sparrowdb-ruby`) converts errors via a catch-all
  `.to_string()` / `Display`, so I don't expect breakage, but I have not
  built or run the Ruby binding itself.
- **Windows.** `std::fs::File::try_lock` is `LockFileEx` there per the std
  docs, and the design doesn't lean on any Unix-only semantics I'm aware of,
  but this was only run on macOS (aarch64-apple-darwin) — no Windows CI leg
  exists in this repo to cross-check against.
- **A true multi-day / many-thousand-cycle stress run.** I ran the
  concurrent-race test dozens of times across the pre-fix and post-fix
  checks (pre-fix: consistent failure; post-fix: consistent pass — never
  flaky either direction in ~15 combined runs), and the full lib-test suite
  4x clean after fixing the one flaky unit test. I have not run it for
  hours or under real production-like load.
- **NFS or other network filesystems.** `flock` semantics on NFS are a
  known trouble spot in general (this is why `SaveLock` in
  `vector_index.rs` doesn't claim NFS support either); this PR doesn't
  claim it either. Only tested on local disk.
- **The out-of-scope repair tool** for already-corrupted databases —
  explicitly not attempted, per the brief. A database already bricked
  before this fix lands is not repaired by it.

Closes #524. Does not fix or attempt the repair-tool question raised in the
issue's peer-review comment — that's tracked as its own, separately-scoped
problem (node ids embed `label_id` in their upper bits, so affected nodes
may be genuinely ambiguous; feasibility unresolved).

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,26 +18,33 @@
 
 ---
 
-> ## ⚠️ Single-process only — concurrent access can permanently destroy your database
+> ## ⚠️ Only one process may hold a database open at a time
 >
-> **Open a SparrowDB database from one process at a time.** Two processes writing concurrently to the
-> same database root can corrupt `catalog.tlv` beyond recovery, after which the database **cannot be
-> opened at all**:
+> **`GraphDb::open` takes an exclusive lock on the database root, process-wide.** A second process that
+> tries to open the same root while another process already has it open gets an immediate, clean error
+> instead of being let in:
 >
 > ```
-> Error: corruption: duplicate label_id 0 in catalog file
+> Error: database locked: another process already has '<path>' open for writing. SparrowDB allows
+> only one open handle per database root at a time — close the other process's connection (or wait
+> for it to exit) and retry.
 > ```
 >
-> Measured on `sparrowdb@0.1.26`: **4 of 5 concurrent runs left the database permanently unopenable.**
-> There is currently **no lock file and no guard at `open()`** — nothing stops a second process from
-> attaching. Sequential access (one process exits, the next opens) is safe.
+> This closes [#524](https://github.com/ryaker/SparrowDB/issues/524): on `sparrowdb@0.1.26` and
+> earlier, there was **no lock file and no guard at `open()`** at all, and two processes writing
+> concurrently to the same root could corrupt `catalog.tlv` beyond recovery — measured at 4 of 5
+> concurrent runs leaving the database permanently unopenable
+> (`Error: corruption: duplicate label_id 0 in catalog file`). If you hit that error on an older
+> version, upgrading does not repair an already-corrupted `catalog.tlv` — it only prevents new
+> corruption from this cause.
 >
-> This is easy to hit by accident: a CLI invoked while a daemon is running, two workers, a cron job
-> overlapping a service, or a health check. The bindings expose only `open()` with **no read-only
-> mode**, so *every* attach is a potential writer.
+> The lock is released automatically when the holding process's handle is dropped or the process
+> exits, including a crash or `SIGKILL` — there is nothing to clean up by hand. Sequential access (one
+> process exits, the next opens) has always been safe and still is.
 >
-> Tracked as [#524](https://github.com/ryaker/SparrowDB/issues/524). Until it is fixed, ensure exactly
-> one process holds a given database root.
+> The bindings still expose only `open()` with **no read-only mode**, so this remains a single-writer
+> constraint rather than true multi-reader support: a second process must wait for the first to close
+> its handle (or exit) before it can attach at all, even just to read.
 
 ---
 
--- a/crates/sparrowdb-common/src/lib.rs
+++ b/crates/sparrowdb-common/src/lib.rs
@@ -70,6 +70,19 @@ pub enum Error {
     /// grows beyond the limit set via `EngineBuilder::with_memory_limit`.
     /// Use a larger limit or restructure the query to reduce fan-out.
     QueryMemoryExceeded,
+    /// `GraphDb::open`/`open_encrypted` could not acquire the exclusive
+    /// cross-process lock on the database root because another process
+    /// already holds it open (#524).
+    ///
+    /// Two processes each deriving `next_label_id` (and other catalog
+    /// counters) from their own in-memory state, with no coordination
+    /// between them, can both allocate the same id and permanently corrupt
+    /// `catalog.tlv` — `GraphDb::open` refuses to hand out a second handle
+    /// to the same root instead of risking that. The lock is released
+    /// automatically when the other process's handle is dropped or the
+    /// process exits (including on crash/kill), so retrying after the
+    /// holder closes its handle succeeds.
+    DatabaseLocked(String),
 }
 
 impl std::fmt::Display for Error {
@@ -107,6 +120,12 @@ impl std::fmt::Display for Error {
             Error::QueryMemoryExceeded => write!(
                 f,
                 "query memory exceeded: BFS frontier exceeded the configured memory limit"
+            ),
+            Error::DatabaseLocked(path) => write!(
+                f,
+                "database locked: another process already has '{path}' open for writing. \
+                 SparrowDB allows only one open handle per database root at a time — close \
+                 the other process's connection (or wait for it to exit) and retry."
             ),
         }
     }
--- a/crates/sparrowdb/src/db.rs
+++ b/crates/sparrowdb/src/db.rs
@@ -71,6 +71,12 @@ impl GraphDb {
     ///
     /// # Errors
     ///
+    /// Returns [`Error::DatabaseLocked`] if another process already has this
+    /// root open — SparrowDB allows only one open handle per database root
+    /// across processes at a time (#524). Retry once the other process
+    /// closes its handle or exits; the lock cannot be stolen or broken, by
+    /// design (see [`process_lock`](crate::process_lock)'s module docs).
+    ///
     /// Besides the usual I/O and WAL-replay failures, this returns
     /// [`Error::Corruption`] when a persisted vector index file exists under
     /// `<path>/vector_indexes/` but cannot be loaded — or when that directory
@@ -163,7 +169,9 @@ impl GraphDb {
     /// fail with [`Error::EncryptionAuthFailed`].
     ///
     /// # Errors
-    /// Returns an error if the directory cannot be created, or
+    /// Returns [`Error::DatabaseLocked`] if another process already has this
+    /// root open — see [`GraphDb::open`]'s identical note (#524). Also
+    /// returns an error if the directory cannot be created, or
     /// [`Error::Corruption`] if a persisted vector index file exists but cannot
     /// be loaded — see [`GraphDb::open`] for why that is fatal.
     pub fn open_encrypted(path: &Path, key: [u8; 32]) -> Result<Self> {
--- a/crates/sparrowdb/src/db.rs
+++ b/crates/sparrowdb/src/db.rs
@@ -86,6 +86,9 @@ impl GraphDb {
     /// unaffected — absent is not the same as damaged.
     pub fn open(path: &Path) -> Result<Self> {
         std::fs::create_dir_all(path)?;
+        // Must be first: refuses a second concurrent handle to this root
+        // before anything below touches the catalog or WAL (#524).
+        let db_lock = crate::process_lock::ProcessLock::acquire(path)?;
         let wal_dir = path.join("wal");
         let wal_writer = WalWriter::open(&wal_dir)?;
         // Replay any WAL records that were fsync'd but not yet written to disk
@@ -147,6 +150,7 @@ impl GraphDb {
                 commits_since_gc: AtomicU64::new(0),
                 vector_indexes,
                 quarantined_vector_writes,
+                _db_lock: db_lock,
             }),
         })
     }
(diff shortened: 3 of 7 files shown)


ryaker:
Merging. **I verified this hands-on rather than reading the diff**, because it's the most severe issue in the repo and because the obvious failure mode of a lock fix is trading one permanent failure for another.

I reproduced my **original** repro — two Node processes racing on one root, the path that bricked the database 4-of-5 times on published 0.1.26 — against a release build of this branch. Not a re-run of the PR's own tests:

```
run 1: Alpha REFUSED (database locked...) | Beta WON  -> DB OPENS, rows=[{"n.id":"Beta"}]
run 2: Beta  REFUSED                      | Alpha WON -> DB OPENS, rows=[{"n.id":"Alpha"}]
... 6/6, exactly one winner each time, database healthy afterwards
```

**The check I most wanted to fail** — a lock that survives a crash would replace "concurrent access destroys the database" with "a crash bricks it until someone deletes a file", which is worse because it looks permanent to the user:

```
holder alive        -> second open refused
SIGKILL the holder  -> next open OPENS OK, rows=[{"n.id":"h"}]   (Drop never runs)
sequential x3       -> all succeed, data accumulates
db.lock present in the root, and its presence alone blocks nothing
```

Kernel-released `flock` on fd close is the right primitive precisely because it makes the stale-lock problem impossible rather than manageable.

**Three things in this PR worth calling out:**

1. **The same-process subtlety was found and handled, not discovered by CI.** `flock` is scoped to the open file description, so a naive implementation would block a second `GraphDb::open` from the *same* process — which `vector_index_durability::a_stale_handle_cannot_silently_revert_another_writers_vectors` deliberately relies on. The process-local registry preserves that while leaving cross-process behaviour untouched. That test would have failed confusingly and been easy to misdiagnose as unrelated flake.
2. **The README banner was rewritten, not left to contradict itself.** #525 shipped a warning saying "no lock file and no guard at `open()`" — now false. A doc asserting a fixed bug is still live sends the next reader chasing nothing, and that has cost this project real time before.
3. **A self-inflicted flaky test was caught and fixed before push** — a registry-wide assertion polluted by sibling unit tests running in parallel. Reported plainly rather than re-run until green.

Review state: claude-review passed; CodeRabbit posted no actionable review. CI green including Benchmarks and the Ruby job that cannot compile locally.

**Known limits, all correctly flagged rather than glossed:** Windows (`LockFileEx`, no CI leg here), NFS/network filesystems (`flock` is a known trouble spot — same caveat `SaveLock` already carries), and no multi-hour soak. None block this: the alternative is a database that concurrent access destroys 4 times in 5.

